PR #8, opened 2017-01-25
removed Xamarin.Android.Tools references to files in obj directory

MihaMarkic:
Xamarin.Android.Tools was contained files from obj directory. That prevented compilation thus I removed them. They were there by mistake anyway, I assume.


tritao:
Yep, mistake. Thanks for the pull.

PR #9, opened 2017-01-25
taking latest visual studio version, not oldest

MihaMarkic:
Currently the oldest VS version is used for compilation instead of newest (latest).


tritao:
Sounds good. I think it might be a good idea to also add an explicit `--vs=version` command line option in the future.

MihaMarkic:
Yes, explicit version argument makes sense.

Issue #11, opened 2017-01-25
Command line option for Visual Studio version


tritao:
As the title says, add a `--vs` CLI option to set the wanted VS version for compilation.

MihaMarkic:
I can take this

MihaMarkic:
Shall we have a finished set of option (like 14.0, 13.0, 12.0 and latest <- default)?

tritao:
I was thinking keying it on VS version (and not the compiler version) as that's how I've historically handled it. so something like `--vs=2015`.

tritao:
Fixed by https://github.com/mono/Embeddinator-4000/commit/e12ece26843579feeeba83b739c7bf69455f9bfe.

PR #12, opened 2017-01-25
added argument for Visual Studio version specification

MihaMarkic:
Defaults to latest, it lets you specify guys from VisualStudioVersion enum (minus VS prefix), so: 2012, 2013, 2015 and Latest.

#11

### DIFF
--- a/binder/Binder.cs
+++ b/binder/Binder.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using CppSharp;
 using CppSharp.Generators;
 
@@ -11,6 +12,7 @@ namespace MonoEmbeddinator4000
         static string Generator;
         static string Platform;
         static string OutputDir;
+        static string VsVersion;
         static List<string> Assemblies;
         static bool CompileCode;
         static bool Verbose;
@@ -21,6 +23,10 @@ namespace MonoEmbeddinator4000
         {
             var showHelp = args.Length == 0;
 
+            string vsVersions = string.Join(", ", 
+                Enum.GetNames(typeof(VisualStudioVersion))
+                .Select(s => s.StartsWith("VS", StringComparison.InvariantCulture) ? s.Substring(2) : s)
+                );
             var optionSet = new Mono.Options.OptionSet() {
                 { "gen=", "target generator (C, C++, Obj-C)", v => Generator = v },
                 { "p|platform=", "target platform (iOS, macOS, Android)", v => Platform = v },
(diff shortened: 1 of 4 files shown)


binder/Binder.cs
```diff
@@ -64,45 +72,60 @@ static void ParseCommandLineArgs(string[] args)
 
         static GeneratorKind ConvertToGeneratorKind(string gen)
         {
-            switch(gen.ToLower())
+            switch (gen.ToLower())
```
tritao: Can you leave these formatting changes out of this commit? We can apply them afterwards in different commit as they are unrelated.

binder/Binder.cs
```diff
@@ -64,45 +72,60 @@ static void ParseCommandLineArgs(string[] args)
...
-                return TargetPlatform.WatchOS;
-            case "tvos":
-                return TargetPlatform.TVOS;
+            if (string.Equals(version, "latest", StringComparison.InvariantCultureIgnoreCase))
+            {
+                return VisualStudioVersion.Latest;
+            }
+            VisualStudioVersion result;
+            if (Enum.TryParse("VS" + version, out result))
+            {
```
tritao: We do not use a block here in our code style since it's single statement.

binder/Binder.cs
```diff
@@ -64,45 +72,60 @@ static void ParseCommandLineArgs(string[] args)
...
-            case "macos":
-                return TargetPlatform.MacOS;
-            case "ios":
-                return TargetPlatform.iOS;
-            case "watchos":
-                return TargetPlatform.WatchOS;
-            case "tvos":
-                return TargetPlatform.TVOS;
+            if (string.Equals(version, "latest", StringComparison.InvariantCultureIgnoreCase))
+            {
```
tritao: Same here.

binder/Compilation.cs
```diff
@@ -398,7 +398,21 @@ void CompileMSVC(IEnumerable<string> files)
...
-            var vsSdk = vsSdks.LastOrDefault();
+            ToolchainVersion vsSdk;
+            if (Options.VsVersion == VisualStudioVersion.Latest)
+            {
+                vsSdk = vsSdks.LastOrDefault();
+            }
+            else
+            {
+                var exactVersion = vsSdks.Where(vs => vs.Version == (float)Options.VsVersion).Cast<ToolchainVersion?>().SingleOrDefault();
+                if (!exactVersion.HasValue)
```
tritao: Same here with unnecessary block.

tritao:
Looks mostly good, can you just apply the minor changes I requested and I will merge it, thanks for contributing.

MihaMarkic:
Yeah, auto formatting. I think it should be good now.
BTW. I'd go with blocks anytime - otherwise you might get heartbleed in no time :)

tritao:
Yeah once in a while some bug could fall through because of it, but I find the code way more readable without them and it also matches the Mono coding guidelines.

PR #25, opened 2017-03-25
[objc] Generate better code for class properties

spouliot:
The following C# code

```
public static class Platform {
	public static bool IsWindows { get { ... } }
}
```

is presently generated (headers) as

```
@interface Platform : NSObject {
    @public MonoEmbedObject* _object;
}

+ (bool)get_IsWindows;
@end
```

which is valid, but awkward looking for objc developers. It is not a class property (like it is in C#) and the `get_` prefix is not expected (nor pretty) for the caller, e.g.

```
NSLog (@"%s", [Platform get_IsWindows] ? "[FAIL]" : "[PASS]");
```

With this PR we generate:

```
@interface Platform : NSObject {
    @public MonoEmbedObject* _object;
}

@property (nonatomic, class, readonly) bool isWindows;
@end
```

which can be called with

```
NSLog (@"%s", [Platform isWindows] ? "[FAIL]" : "[PASS]");
```

### DIFF
--- a/binder/Generators/ObjC/ObjCGenerator.cs
+++ b/binder/Generators/ObjC/ObjCGenerator.cs
@@ -28,15 +28,66 @@ namespace MonoEmbeddinator4000.Generators
 
     public static class ObjCExtensions
     {
-        public static void GenerateObjCMethodSignature(this CCodeGenerator gen,
-            Method method)
+        public static bool IsProperty (this Method method)
         {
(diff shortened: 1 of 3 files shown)


tritao:
Looks like my previous comment on this PR didn't go through...

This is a nice feature but I think it would be best to use a different approach.
We already have a `Property` AST node that can be used for generating properties, that way we do not need to be trying to figure out if a method is a property or not.

I'll rework your code a little bit and merge this manually.

PR #38, opened 2017-03-31
[objc] Add support for default .ctor/init and namespaces

spouliot:


objcgen/objcgenerator.cs
```diff
@@ -102,36 +103,75 @@ protected override void Generate (Type t)
...
+
+					implementation.WriteLine ("- (instancetype)init");
+					implementation.WriteLine ("{");
+					implementation.WriteLine ($"\tconst char __method_name [] = \"{t.FullName}:.ctor()\";");
+					implementation.WriteLine ("\tstatic MonoMethod* __method = nil;");
+					implementation.WriteLine ("\tif (!__method) {");
+					implementation.WriteLine ($"\t\t__lookup_class_{managed_name} ();");
+					implementation.WriteLine ($"\t\t__method = mono_embeddinator_lookup_method (__method_name, {managed_name}_class);");
+					implementation.WriteLine ("\t}");
+					implementation.WriteLine ("\tif (self = [super init]) {");
```
rolfbjarne: This won't work for managed classes that inherit from NSObject (in which case we should just call the managed constructor instead of `[super init]`).

spouliot: I can add a TODO but I suspect quite a few changes to support XI / XM types.

rolfbjarne: Yeah, it's fine for now.

PR #39, opened 2017-04-01
[objc] Add support for -o (output directory) and -v (verbosity) and add some error codes

spouliot:
* EM0000 is now shown for unhandled exceptions - asking for a bug report
  to be submitted;

* EM0001 is reported when the output directory cannot be found;

* EM1000 is reported when the generator throws a NotImplementedException,
  no bug report is required, but contributions are welcome :)

### DIFF
--- a/objcgen/driver.cs
+++ b/objcgen/driver.cs
@@ -38,7 +38,9 @@ namespace Embeddinator {
 			var action = Action.None;
 
 			var os = new OptionSet {
+				{ "o|out|outdir=", "output directory", v => OutputDirectory = v },
 				{ "h|?|help", "Displays the help", v => action = Action.Help },
+				{ "v|verbose", "generates diagnostic verbose output", v => ErrorHelper.Verbosity++ },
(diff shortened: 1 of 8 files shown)


objcgen/driver.cs
```diff
@@ -53,20 +55,35 @@ public static int Main2 (string [] args)
...
 				throw ErrorHelper.CreateError (99, "Internal error: invalid action {0}. Please file a bug report with a test case (https://github.com/mono/Embeddinator-4000/issues)", action);
 			}
 		}
 
+		static string outputDirectory = ".";
+
+		public static string OutputDirectory {
+			get { return outputDirectory; }
+			set {
+				if (!Directory.Exists (value))
```
rolfbjarne: Can't we just create the directory if it doesn't exist?

spouliot: It's not typo friendly (e.g. "../../../test/bin/debug" that has an extra, or missing, "../") and they are more common from a (currently) CLI tool. OTOH we can warn when we create.

PR #42, opened 2017-04-01
[objc] Add xcode support for objc-cli to build the libmanaged.dylib and it's unit tests

spouliot:
This allows debugging the generated code with Xcode as well as
running the native unit tests with XCTest (which might not be the
best tool but it's good to support).

This commit requires a addition to mono_embeddinator so we can
override where the assemblies are loaded from - we cannot assume
they will be with the executable (e.g. plugin) and we hit this
case with Xcode unit tests.

### DIFF
--- a/support/mono_embeddinator.c
+++ b/support/mono_embeddinator.c
@@ -119,8 +119,17 @@ int mono_embeddinator_destroy(mono_embeddinator_context_t* ctx)
     return true;
 }
 
+static GString* path_override = NULL;
+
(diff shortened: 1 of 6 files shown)


spouliot:
Just realized that `Makefile` files were not added (due to the .gitignore at the top of the repo).

`cd libmanaged && xcodebuild test -scheme Tests`

will build the lib, using Xcode, and run the ObjC tests. From what I've seen Jenkins has support (plugin) to run Xcode unit tests, so it should be possible to add it to the bots without custom logic.

PR #48, opened 2017-04-03
[objc] Setters don't have return values.

rolfbjarne:


rolfbjarne:
Looks like this is already fixed.

PR #55, opened 2017-04-03
[objc][tests] Remove the 'libmanaged' target from the Xcode cli test project.

rolfbjarne:
Remove the 'libmanaged' target from the Xcode test project, and instead link
with the libmanaged.dylib that the embeddinator compiles. This ensures we're
not compiling with different compiler flags in the Xcode project.

This also requires adding an rpath to the built library so that the test
executable can find it.

Also add makefile target to run the Xcode test project from the command line.


rolfbjarne:
build

PR #58, opened 2017-04-04
[objc] Add memory leak tests, and fix a few issues those found.

rolfbjarne:


tests/objc-cli/test-managed.m
```diff
@@ -22,17 +22,21 @@ int main ()
 		NSLog (@"%s instance property getter", [query answer] == 42 ? "[PASS]" : "[FAIL]");
 		query.answer = 911;
 		NSLog (@"%s instance property setter", [query answer] == 911 ? "[PASS]" : "[FAIL]");
+		[query release];
```
spouliot: should we not enforce ARC instead ?

rolfbjarne: I don't think we can _enforce_ ARC in code that consumes the bindings, but this is code we control ourselves, so we can switch to ARC if we want to.

rolfbjarne:
build

tests/objc-cli/test-managed.m
spouliot: Seems that we can [1] and I think we should - just to cut down on support issues and the test matrix.

[1] http://tewha.net/2012/09/make-your-library-enforce-arc/

rolfbjarne: You mean we should put this in bindings.h?

```c
#if !__has_feature(objc_arc)
#error Embeddinator code must be built with ARC.
#endif
```

IMHO that's a bit harsh, but I guess we can always remove it later if it becomes a problem.

spouliot: it may be harsh but I rather support less now (and _possibly_ more later) and focus on the called (not the caller) code :)

rolfbjarne:
I have modified this quite a bit, and I'll push a new PR.

PR #62, opened 2017-04-05
[objc][generator] Fix subclassing when both types are bound

spouliot:
E.g. `public class SuperUnique : Unique {}`

This raise an interesting issue (to be solved later) since ObjC and .NET
semantics differs for init*/.ctor. A test case (with a FIXME) is added.

and I now recall why headers were a bad idea:

./bindings.h:75:39: error: attempting to use the forward class 'Constructors_Unique' as superclass of 'Constructors_SuperUnique'

so the ordering of forwarders had to be altered a bit.

### DIFF
--- a/objcgen/objcgenerator.cs
+++ b/objcgen/objcgenerator.cs
@@ -17,7 +17,8 @@ namespace ObjC {
 		static TextWriter implementation = new StringWriter ();
 
 		List<Type> types = new List<Type> ();
-		Dictionary<Type,List<ConstructorInfo>> ctors = new Dictionary<Type,List<ConstructorInfo>> ();
+		Dictionary<Type, List<ConstructorInfo>> ctors = new Dictionary<Type, List<ConstructorInfo>> ();
+		Dictionary<Type, List<PropertyInfo>> properties = new Dictionary<Type, List<PropertyInfo>> ();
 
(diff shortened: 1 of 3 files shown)


spouliot:
failed with mono 4.8.0 since we use .pdb

```
error: /Users/builder/jenkins/workspace/embeddinator-4000/tests/managed/bin/Debug/managed.pdb: No such file or directory
```

Maybe we should require a minimum of mono 5.0 ?

spouliot:
build

objcgen/objcgenerator.cs
```diff
@@ -130,7 +137,9 @@ protected override void Generate (Type t)
 			headers.WriteLine ();
 			headers.WriteLine ($"// {t.AssemblyQualifiedName}");
 			headers.WriteLine ($"@interface {native_name} : {GetTypeName (t.BaseType)} {{");
-			headers.WriteLine ("\tMonoEmbedObject* _object;");
+			// our internal field is only needed once in the type hierarchy
```
rolfbjarne: We only need the dealloc method once too.

rolfbjarne: Actually NVM, we're not generating the dealloc method in master yet :smile:

PR #69, opened 2017-04-06
[objc] Use enums instead of string values for platform/target/compilationtarget, and add initial scaffolding for iOS/watchOS/tvOS + framework support.

rolfbjarne:


tests/objcgentest/Asserts.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using NUnit.Framework;
+
+using Embeddinator;
+
+namespace DriverTest
+{
+	public static class Asserts
+	{
+		public static void ThrowsEmbeddinatorException (int code, Action action)
```
spouliot: 👍

PR #70, opened 2017-04-06
[tests] Add pre-existing tests to managed test suite

tritao:


tritao:
The idea here is to share the managed.dll test project between both generators. I'll be adding some more commits to the PR to port the C and Java test runner to `managed.dll`.

spouliot:
@tritao I'm fine with that goal :) but the `objc` branch must remain green and unused code/tests needs to be identified as such.

tritao:
Makes sense of course, I'll fix it the way you suggested.

tritao:
Updated PR with feedback.

tests/managed/types.cs
```diff
@@ -80,3 +80,40 @@ public static class Type_String
...
+	public sbyte PassAndReturnsSByte (sbyte v) { return v; }
+	public byte PassAndReturnsByte (byte v) { return v; }
+	public short PassAndReturnsShort (short v) { return v; }
+	public ushort PassAndReturnsUShort (ushort v) { return v; }
+	public int PassAndReturnsInt (int v) { return v; }
+	public uint PassAndReturnsUInt (uint v) { return v; }
+	public long PassAndReturnsLong (long v) { return v; }
+	public ulong PassAndReturnsULong (ulong v) { return v; }
+	public char PassAndReturnsChar (char v) { return v; }
+	public string PassAndReturnsString (string v) { return v; }
```
spouliot: all of the above should be working, they just need to be added in the Xcode unit tests (to confirm and avoid regressions)

tests/managed/arrays.cs
```diff
@@ -0,0 +1,27 @@
+#if NON_OBJC_SUPPORTED_TESTS
```
spouliot: maybe `!TARGET_OBJC` ? or something similar.  We'll likely have some cases where other generator or platforms needs customization (temporary or permanent). @rolfbjarne ?

tritao: Yeah I like that better..

tests/managed/types.cs
tritao: OK I can try adding them to Xcode tests.

tests/managed/arrays.cs
rolfbjarne: Yeah, I like `TARGET_OBJC` better too.

spouliot: sold! :)

tritao:
Not needed anymore since ObjC branch can now deal with everything besides arrays, I'll integrate those tests once array support is added.

PR #88, opened 2017-04-10
[objc][doc] Update examples and add Objective-C coding guidelines document

VincentDondain:


VincentDondain:
Same as https://github.com/mono/Embeddinator-4000/pull/87 but against `objc` branch.

Issue #82, opened 2017-04-07
[objc][perf] Use metadata tokens in some cases


spouliot:
Right now the type and method lookups are done using strings. 

```
	const char __method_name [] = "First.Second.Third.ClassWithNestedNamespace:ToString()";
	static MonoMethod* __method = nil;
	if (!__method) {
		__lookup_class_First_Second_Third_ClassWithNestedNamespace ();
		__method = mono_embeddinator_lookup_method (__method_name, First_Second_Third_ClassWithNestedNamespace_class);
	}
```

That has a nice advantage (that we want to exploit) as it allow swapping a .dll (if the API is identical) on some platforms (e.g. won't work with iOS/AOT). This can be very useful in **debug** mode.

However we could use the metadata tokens when building **release** code. That should (benchmarking needed) provide a performance boost (on first call) and reduce the generated (and final) code side (since many strings can go away).

spouliot:
implemented with PR #91

PR #91, opened 2017-04-10
[objc][generator] Use metadata token lookup when building release

spouliot:
where release is non-debug.

https://github.com/mono/Embeddinator-4000/issues/82


spouliot:
yup, I'll wait for the `-o` (Output) fix to be merged and update the PR :)

PR #97, opened 2017-04-11
[objc] Add support fat and static libraries for all platforms.

rolfbjarne:
This also:
* Fixes the -o support (by actually using the output directory)
* Revert the change to make the `_object` private (in the implementation
  only), because this doesn't compile when building for the legacy Objective-C
  runtime (32-bit macOS).

### DIFF
--- a/objcgen/driver.cs
+++ b/objcgen/driver.cs
@@ -228,59 +228,263 @@ namespace Embeddinator {
 			return 0;
 		}
 
+		class BuildInfo
+		{
+			public string Sdk;
+			public string [] Architectures;
+			public string SdkName; // used in -m{SdkName}-version-min
+			public string MinVersion;
+			public string XamariniOSSDK;
+			public string CompilerFlags;
+			public string LinkerFlags;
+		}
+
 		static int Compile ()
 		{
 			Console.WriteLine ("Compiling binding code...");
 
+			BuildInfo [] build_infos;
+
 			switch (Platform) {
 			case Platform.macOS:
+				build_infos = new BuildInfo [] {
+					new BuildInfo { Sdk = "MacOSX", Architectures = new string [] { "i386", "x86_64" }, SdkName = "macosx", MinVersion = "10.7" },
+				};
 				break;
 			case Platform.iOS:
-			case Platform.watchOS:
+				build_infos = new BuildInfo [] {
+					new BuildInfo { Sdk = "iPhoneOS", Architectures = new string [] { "armv7", "armv7s", "arm64" }, SdkName = "iphoneos", MinVersion = "8.0", XamariniOSSDK = "MonoTouch.iphoneos.sdk" },
+					new BuildInfo { Sdk = "iPhoneSimulator", Architectures = new string [] { "i386", "x86_64" }, SdkName = "ios-simulator", MinVersion = "8.0", XamariniOSSDK = "MonoTouch.iphonesimulator.sdk" },
+				};
+				break;
 			case Platform.tvOS:
-				throw new NotImplementedException ($"platform={Platform}");
+				build_infos = new BuildInfo [] {
+					new BuildInfo { Sdk = "AppleTVOS", Architectures = new string [] { "arm64" }, SdkName = "tvos", MinVersion = "9.0", XamariniOSSDK = "Xamarin.AppleTVOS.sdk", CompilerFlags = "-fembed-bitcode", LinkerFlags = "-fembed-bitcode" },
+					new BuildInfo { Sdk = "AppleTVSimulator", Architectures = new string [] { "x86_64" }, SdkName = "tvos-simulator", MinVersion = "9.0", XamariniOSSDK = "Xamarin.AppleTVSimulator.sdk" },
+				};
+				break;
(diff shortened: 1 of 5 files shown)


rolfbjarne:
Don't merge because it depends on PR #96.

tests/objc-cli/Makefile
```diff
@@ -52,3 +52,13 @@ test-xctest-leaks: ../leaktest/bin/Debug/leaktest.exe $(MANAGED_DLL) libmanaged.
 
 libLeakCheckAtExit.dylib: leak-at-exit.c
 	clang -arch i386 -arch x86_64 -shared $< -o $@
+
+test-static: test-static-macos test-static-ios test-static-tvos test-static-watchos
+
+test-static-%: managed.dll $(OBJC_GEN)
+	/Library/Frameworks/Mono.framework/Versions/Current/Commands/mono --debug $(OBJC_GEN) --debug managed.dll -c --outdir=build/$@-temp-dir --target=staticlibrary --platform=$*
+
+test-dynamic: test-dynamic-macos test-dynamic-ios test-dynamic-tvos test-dynamic-watchos
```
spouliot: minor: we should consistently use either `shared` or `dynamic` across code (and docs)

objcgen/objcgenerator.cs
```diff
@@ -151,15 +151,17 @@ protected override void Generate (Type t)
 			var native_name = GetTypeName (t);
 			headers.WriteLine ();
 			headers.WriteLine ($"// {t.AssemblyQualifiedName}");
-			headers.WriteLine ($"@interface {native_name} : {GetTypeName (t.BaseType)}");
```
spouliot: what's the exact build that failed ? Apple has been removing this pattern from headers for a while - but maybe I never noticed they did not touch the macOS 32bits ones ?

tests/objc-cli/Makefile
```diff
@@ -52,3 +52,13 @@ test-xctest-leaks: ../leaktest/bin/Debug/leaktest.exe $(MANAGED_DLL) libmanaged.
 
 libLeakCheckAtExit.dylib: leak-at-exit.c
 	clang -arch i386 -arch x86_64 -shared $< -o $@
+
+test-static: test-static-macos test-static-ios test-static-tvos test-static-watchos
+
+test-static-%: managed.dll $(OBJC_GEN)
+	/Library/Frameworks/Mono.framework/Versions/Current/Commands/mono --debug $(OBJC_GEN) --debug managed.dll -c --outdir=build/$@-temp-dir --target=staticlibrary --platform=$*
```
spouliot: I assume the execution part is incoming in a future commit ?

objcgen/objcgenerator.cs
rolfbjarne: Yes, more changes are needed for that, but I wanted to get this in since other people were running into issues I'd already fixed here (-o didn't work).

tests/objc-cli/Makefile
rolfbjarne: maybe `dylib` and `framework` when that distinction is important (in this case it is, but not yet), since both `shared` and `dynamic` are technically correct for both of them.

And then also choose between `shared` and `dynamic` of course.

tests/objc-cli/Makefile
rolfbjarne: Yes, more changes are needed for that, but I wanted to get this in since other people were running into issues I'd already fixed here (-o didn't work).

objcgen/objcgenerator.cs
rolfbjarne: This is how it fails:

```
test.m:7:20: error: inconsistent number of instance variables specified
@public NSObject * _object;
                   ^
1 error generated.
```

Test case:
```objective-c
#include <Foundation/Foundation.h>

@interface X : NSObject
@end

@implementation X {
@public NSObject * _object;
}
@end

int main ()
{
	return 0;
}
```

built using:
```shell
clang test.m -framework Foundation -arch i386
```

See also: http://stackoverflow.com/a/15174088/183422

spouliot: ok 😢 makes me wonder if we should support 32bits macOS... 
seriously maybe the headers could be (later) made more modern for iOS/tvOS/watchOS

objcgen/driver.cs
```diff
@@ -228,59 +228,263 @@ static int Generate (List<string> args)
...
+					var archOutputDirectory = Path.Combine (OutputDirectory, arch);
+					Directory.CreateDirectory (archOutputDirectory);
+
+					var common_options = new StringBuilder ("clang ");
+					if (Debug)
+						common_options.Append ("-g -O0 ");
+					common_options.Append ("-fobjc-arc ");
+					common_options.Append ("-ObjC ");
+					common_options.Append ($"-arch {arch} ");
+					common_options.Append ($"-isysroot /Applications/Xcode.app/Contents/Developer/Platforms/{build_info.Sdk}.platform/Developer/SDKs/{build_info.Sdk}.sdk ");
```
spouliot: that cause an error to me - I don't have that directory (or a symlink)

PR #98, opened 2017-04-11
[objc] Get the Xcode location using 'xcode-select -p' to use for our clang compiler flags.

rolfbjarne:


rolfbjarne:
According to `xcode-select`'s man page: "xcode-select controls the location of the developer directory used by xcrun(1)...", so the behavior should be identical.

A simple test also shows that `xcode-select` honors `DEVELOPER_DIR`:

```shell
$ DEVELOPER_DIR=/Applications/Xcode.app xcode-select -p
/Applications/Xcode.app/Contents/Developer
$ xcode-select -p
/Applications/Xcode83.app/Contents/Developer

```

PR #101, opened 2017-04-11
[objc] Disable the `new` selector when no `init` is available

dalexsoto:
We currently disable the `init` ctor when is not needed i.e. static
classes in .NET but we should also disable the `new` static selector
since you could do `[foo new]` and get an instance.


dalexsoto:
@spouliot  IIRC yes, this is needed, Xcode will complain about `init' but won't about 'new'

newobjc

dalexsoto:
just tested, this will fail until runtime with

>*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'The -init method is not available on HKUnit'

So we better fail early and fast :)

PR #108, opened 2017-04-12
[tests][objc] Build debug tests into a subdirectory as well.

rolfbjarne:
This cleans up the current directory quite a bit.

Also modify the cli/perf tests a little bit to be able to run the
debug/release versions in parallel.


rolfbjarne:
Yes, it's running perf in parallel. I think you'll have to move perf tests to a different directory if you don't want anything else running at the same time.

PR #129, opened 2017-04-16
[objc] Add basic (readonly) support for fields

spouliot:
We cannot access them directly but we can provide (ObjC) property
wrappers around them, making them usable from ObjC code.

### DIFF
--- a/objcgen/objcgenerator.cs
+++ b/objcgen/objcgenerator.cs
@@ -266,20 +266,25 @@ namespace ObjC {
 				implementation.WriteLine ();
 			}
 
-			headers.WriteLine ();
 			List<PropertyInfo> props;
 			if (properties.TryGetValue (t, out props)) {
+				headers.WriteLine ();
 				foreach (var pi in props)
 					Generate (pi);
-				headers.WriteLine ();
 			}
 
-			headers.WriteLine ();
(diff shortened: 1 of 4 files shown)


objcgen/objcgenerator.cs
```diff
@@ -371,6 +376,59 @@ protected override void Generate (PropertyInfo pi)
...
+		{
+			bool read_only = ((fi.Attributes & FieldAttributes.InitOnly) == FieldAttributes.InitOnly);
+
+			headers.Write ("@property (nonatomic");
+			if (fi.IsStatic)
+				headers.Write (", class");
+			if (read_only)
+				headers.Write (", readonly");
+			else
+				headers.Write (", readwrite");
```
dalexsoto: This is the default `readwrite`, it does not harm to have it but it is a little uncommon to see in ObjC do we really want to have it?

spouliot: It's already generated for _normal_ properties - but we can PR their removal (both) since, being the default, it has no/low value to ObjC developers.

spouliot:
@dalexsoto afaik we're following the ARC defaults (and like `readwrite` we should not write defaults), found any other case ? If so they need to be fixed, but I don't think it should be in this PR

PR #132, opened 2017-04-17
[objc] Removes `readwrite` in objc properties since it is the default

dalexsoto:
Adding `readwrite` to objective-c properties adds no value since it
is the default setting.

Removal was suggested in https://github.com/mono/Embeddinator-4000/pull/129#issuecomment-294491293


dalexsoto:
Ahh wrong branch

PR #149, opened 2017-04-20
[objc] Add support for compiling to frameworks.

rolfbjarne:
Add support for compiling to frameworks.

I've only tested iOS frameworks, tvOS and watchOS may or may not work.

macOS is not supported (and will show errors).

### DIFF
--- a/objcgen/driver.cs
+++ b/objcgen/driver.cs
@@ -502,6 +502,11 @@ namespace Embeddinator {
 						Directory.CreateDirectory (headers);
 						File.Copy (Path.Combine (OutputDirectory, "embeddinator.h"), Path.Combine (headers, "embeddinator.h"), true);
 						File.Copy (Path.Combine (OutputDirectory, "bindings.h"), Path.Combine (headers, $"bindings.h"), true);
+						// Create an umbrella header for the everything in the framework.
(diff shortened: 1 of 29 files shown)


objcgen/driver.cs
```diff
@@ -350,6 +346,25 @@ public int Compile ()
...
+			/*
+			 * For static libraries we
+			 * * Compile all source files to .o files, per architecture.
+			 * * Then we archive .o files into per-sdk .a files, then we archive all .o files into a global .a file. So we end up with one .a for device, one for simulator, and one for both device and simulator.
+			 * * Then we dsym the global .a file
+			 * 
+			 * For dynamic libraries we
+			 * * Compile all source files to .o files, per architecture.
+			 * * Then we link all .o files per architecture into a .dylib.
+			 * * Then we lipo .dylib files into per-sdk fat .dylib, then we lipo all .dylbi into a global .dylib file. So we end up with one fat .dylib for device, one fat .dylib for simulator, and one very fat .dylib for both simulator and device.
```
dalexsoto: Super low: Typo 
> we lipo all .dylbi

spouliot:
@rolfbjarne I'm fine with disabling the build/test of the C generator (until fixed).

PR #151, opened 2017-04-20
[objc] Fix clang warnings about nullability

spouliot:
In general we don't encourage `null` since it's likely to get managed
`ArgumentNullException` which cannot be well handled on the ObjC side.
That's covered by:

> NS_ASSUME_NONNULL_BEGIN

Right now we lack any metadata to make this more precise :-(. However,
when possible, we can put specific annotations on some API, e.g.

1. Any managed exception in a `.ctor` (or the `.cctor`) will make the
corresponding `init*` method return `nil`, so we always mark them as
`nullable`.

2. `ref` and `out` on reference types can be `nullable`

### DIFF
--- a/docs/Limitations.md
+++ b/docs/Limitations.md
@@ -13,3 +13,8 @@ It is not possible to have two mono runtimes co-exists inside the same applicati
 
 ## ObjC generated code
 
+### Nullability
+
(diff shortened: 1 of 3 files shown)


spouliot:
Travis timed out

PR #157, opened 2017-04-21
[objc] Add support for default values on .ctor/method parameters

spouliot:
note: incomplete, non-integral types needs more love

### DIFF
--- a/objcgen/objcgenerator-processor.cs
+++ b/objcgen/objcgenerator-processor.cs
@@ -86,7 +86,7 @@ namespace ObjC {
 						delayed.Add (ErrorHelper.CreateWarning (1020, $"Constructor `{ctor}` is not generated because of parameter type `{pt}` is not supported."));
 						pcheck = false;
 					} else if (p.HasDefaultValue) {
-						delayed.Add (ErrorHelper.CreateWarning (1021, $"Constructor `{ctor}` parameter `{p.Name}` has a default value that is not supported."));
+						members_with_default_values.Add (ctor);
 					}
 				}
 				if (!pcheck)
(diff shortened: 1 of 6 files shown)


spouliot:
@rolfbjarne I mentioned that on slack, I never realized (or forgot) the C# compiler allowed this :(

Right now [1] the build fails in such case since we do not have support to detect conflicts. Ideally we'll be able to query if the default `.ctor` exists but it needs to be more general than this (another identical signature could also be produced).

[1] this is coming with PR #156 but it will require more changes - both in existing code, like `IComparable`, `GetHashCode`/`Equals`, and the PR. E.g. need to register the _intent_ of adding `X` before the duplication checks are done.

rolfbjarne:
BTW this is not limited to constructors, all methods can overload each other like this.

From what I can see you're generating every possibles signature that results from removing default parameters one by one from the right.

So for `void X (int a = 0, int b = 0)`

You bind as if these methods were there:

```
void X (int a, int b);
void X (int a);
void X ();
```

In this particular case (I've just glanced at PR #156, so I don't know what it's capable of), I think we should just not generate bindings for the signatures that already exists in the type.

So if the user has:

```
void X (int a);
void X (int a = 0, int b = 0);
```

We'll generate:

```
void X (int a); // for the first method
void X (int a, int b); // for the second method
void X (); // for the second method
```

spouliot:
@rolfbjarne yes, it's method too. Right now having:

```
void X (int a);
void X (long a);
```

fails too, without being related to default values. That's why PR #156 needs to be addressed quickly and it's only a start, i.e. address the later problem.

Issue #104, opened 2017-04-12
[objc] Add error/warning tests for things we can't execute


dalexsoto:
From @spouliot https://github.com/mono/Embeddinator-4000/pull/103#discussion_r111054181

> We will need to do some error/warning tests (like touch tests) to ensure we don't regress things we can't execute. OTOH that can be done in a separate PR.

spouliot:
PR #203 was merged

PR #158, opened 2017-04-21
[objc][tests] Add objcgen warning and error tests. Fixes issue #104

dalexsoto:
The objective is to test warning and error conditions that you normally can't using unit tests.

## Testing objcgen.exe

Currently, we have 2 test templates inside `Makefile` for testing *objcgen.exe* error and warning codes.

* **FindGenErrWarnCodeTemplate**: This is used to verify that error and warning codes are actually printed in *objcgen.exe* log output.
* **DoNotFindGenErrWarnCodeTemplate**: This is used to verify that error and warning are not printed in *objcgen.exe* log output.

In order to add additional tests create a new folder that includes the C# project with its files and add the following to the makefile:

```makefile
$(eval $(call {Template},{TEST},{csprojName},{Debug|Release},{[warn|error]code}))
```

For example:

```makefile
$(eval $(call FindGenErrWarnCodeTemplate,GenericsTest,GenericsLib,Debug,1010))
```

Also, do not forget to add your folder name into `TESTS` variable inside Makefile.

## Testing error conditions when using binding headers produced by objcgen.exe

Currently, we have 1 test template inside Makefile for testing binding headers.

* **FindXcodeErrWarnCodeTemplate**: This is used to verify that an error or a warning is printed in the log output when building an *xcodeproj*.

In order to add additional tests create a new folder that includes the C# project with its files that *objcgen.exe* will use as input, an Xcode project that uses the output headers of *objcgen.exe* and add the following to the makefile:

```makefile
$(eval $(call {Template},{TEST},{csprojName},{xcodeprojName},{xcodeTarget},{Debug|Release},{expected[warn|error]message}))
```

For example:

```makefile
$(eval $(call FindXcodeErrWarnCodeTemplate,NoInitInSubclassTest,ConstructorsLib,NoInitInSubclassTest,NoInitInSubclassTest,Release,"error: 'initWithId:' is unavailable"))
```

The `{expected[warn|error]message}` is given as a `grep` argument if it does not find it the test will fail. Also, do not forget to add your folder name into `TESTS` variable inside Makefile.

## Custom tests

If none of the above templates matches your testing needs feel free to add a makefile target with your own testing logic.

### DIFF
--- a/tests/objcgenwarnerrtests/EnumsTest/EnumsLib.csproj
+++ b/tests/objcgenwarnerrtests/EnumsTest/EnumsLib.csproj
@@ -1,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{4F4AC361-08F3-4566-839A-A6B3E740F81A}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <RootNamespace>EnumsLib</RootNamespace>
+    <AssemblyName>EnumsLib</AssemblyName>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <OutputPath>bin\Debug</OutputPath>
+    <DefineConstants>DEBUG;</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <ConsolePause>false</ConsolePause>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+    <Optimize>true</Optimize>
+    <OutputPath>bin\Release</OutputPath>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <ConsolePause>false</ConsolePause>
+  </PropertyGroup>
+  <ItemGroup>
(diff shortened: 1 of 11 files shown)


dalexsoto:
Build

dalexsoto:
Jenkins did not run this tests on last buld because it did not know about them now it does

dalexsoto:
```
14:35:01 + make -C tests/objcgenwarnerrtests all
14:35:01 GenericsTest
14:35:05 EnumsTest
14:35:09 NoInitInSubclassTest
```

rolfbjarne:
I'm not quite sure we need this.

We can already test generator errors in unit tests: https://github.com/mono/Embeddinator-4000/blob/5754f599918ecb456f41ca2c663e8c822e93c517/tests/objcgentest/DriverTest.cs#L155

Warnings is a bit more complicated, but I think we can just do something like this:

```
    var stdout = Console.Out;
    var stderr = Console.Error;
    var txtWriter = new TextWriter ();
    Console.SetOut (txtWriter);
    Console.SetError (txtWriter);
    /* do something that causes warnings */
    /* process txtWriter contents into error/warning messages, and assert the values */
    Console.SetOut (stdout);
    Console.SetError (stderr);
```

That's a lot simpler that writing the logic in Makefiles, and additionally you can run single tests using the test pad in Xamarin Studio.

The downside is that NUnit doesn't support running tests in parallel (not yet at least), so it might run a little bit slower (but on the other hand this doesn't require launching a new process for every error/warning message).

spouliot:
Good idea, it's worth a try as it will require less maintenance (and we have more great people for developing in C# than Makefiles).

dalexsoto:
@rolfbjarne I think I got something working for warnings using `objcgentest`. Now about testing errors in Xcode projects consuming our stuff like https://github.com/mono/Embeddinator-4000/pull/158/files#diff-7119fd8833ac63ca79b4bd69ada72f40R29 do you have a better idea than a makefile? 😃

rolfbjarne:
@dalexsoto you can still just run the commands from a unit test.

A very simplified version would be something like this:

```
[Test]
[TestCase ("NoInitInSubclassTest", "ConstructorsLib", "NoInitInSubclassTest", "NoInitInSubclassTest", "Release", "error: 'initWithId:' is unavailable")]
void MyTest (string projectDir, string project, string xcodeProject, string xcodeTarget, string config, string expectedError)
{
    Process.Start (MONOMSBUILD, $"/p:Configuration={config} {projectDir}/{project}.csproj /p:OutputPath=DllOutput").WaitForExit ();
    Process.Start (MONO, ...).WaitForExit ();
    Process.Start ("xcodebuild", "...").WaitForExit ();
}
```

dalexsoto:
Closed in favor of PR #203

Issue #123, opened 2017-04-14
[objc] Map `GetHashCode` and `Equals` to `NSObject` protocol `hash` and `isEqual:`


spouliot:
The semantics are identical.

`@property(readonly) NSUInteger hash;`

* `hash` returns an `NSUInteger` so it differs in 32/64 bits, while .NET always return a 32bits integer;

https://developer.apple.com/reference/objectivec/1418956-nsobject/1418859-hash?language=objc

`- (BOOL)isEqual:(id)object;`

* a non-bound `NSObject` (i.e. does not have a managed peer) must not be sent to the mono runtime.

https://developer.apple.com/reference/objectivec/1418956-nsobject/1418795-isequal?language=objc

spouliot:
That would remove warnings like:

> warning EM1031: Method `Boolean Equals(System.Object)` is not generated because of parameter type `System.Object` is not supported.

timrisi:
I'll look into this one

timrisi:
https://github.com/mono/Embeddinator-4000/pull/159

timrisi:
Merged

PR #159, opened 2017-04-22
[objc] Add support for overriding Equals and GetHashCode. Fixes issue #123

timrisi:


objcgen/HashHelper.cs
```diff
@@ -0,0 +1,24 @@
...
+
+		public void WriteImplementation ()
+		{
+			BeginImplementation ();
+			WriteMethodLookup ();
+			WriteInvoke ("nil");
+
+			implementation.WriteLine ("\tvoid* __unbox = mono_object_unbox (__result);");
+			implementation.WriteLine ("\treturn (NSUInteger)(*((int*)__unbox));");
+			//implementation.WriteLine ("\treturn (NSUInteger)result;");
```
spouliot: ^ you can remove the commented code

objcgen/objcgenerator.cs
```diff
@@ -367,6 +367,35 @@ protected override void Generate (Type t)
...
+				builder.WriteHeaders ();
+				builder.WriteImplementation ();
+
+
+				GenerateGetGCHandle ();
+			}
+
+			if (hashes.TryGetValue (t, out m)) {
+				var builder = new HashHelper (headers, implementation) {
+					ObjCSignature = $"hash",
```
spouliot: no need for `$` here - even if, _I suspect_, the C# compiler won't call `Format` unless needed

objcgen/objcgenerator.cs
```diff
@@ -367,6 +367,35 @@ protected override void Generate (Type t)
 				builder.WriteImplementation ();
 			}
 
+			if (equals.TryGetValue (t, out m)) {
+				var builder = new EqualsHelper (headers, implementation) {
+					ObjCSignature = $"isEqual:(id)other",
```
spouliot: no need for `$` here - even if, _I suspect_, the C# compiler won't call `Format` unless needed

objcgen/EqualsHelper.cs
```diff
@@ -0,0 +1,46 @@
...
+		public EqualsHelper (TextWriter headers, TextWriter implementation) : base (headers, implementation)
+		{
+			MonoSignature = "Equals (object)";
+			ReturnType = "bool";
+		}
+
+		public void WriteImplementation ()
+		{
+			BeginImplementation ();
+			implementation.WriteLine ("\tif (![other respondsToSelector: @selector (xamarinGetGCHandle)])");
```
rolfbjarne: What if `other` is `NULL`?

timrisi: respondsToSelector will return nil and it will return false.  Double checked in an iOS app with:

```
NSObject *object;
    if (![object respondsToSelector: @selector(foo)])
        NSLog (@"Responds");
```

NSObject is nil and the selector doesn't exist, it hits a breakpoint set on the NSLog

rolfbjarne: Can you add this to the tests?

timrisi: Already there, first test compares it to nil:

XCTAssertFalse ([c1 isEqual:nil], "equals nil");

spouliot:
@rolfbjarne I think @timrisi answered your question above

objcgen/EqualsHelper.cs
rolfbjarne: 👍

PR #184, opened 2017-04-25
[objc] Add ClassHelper and use Indent for remaining tab support

spouliot:
ClassHelper will the addition of @protocol without duplicating too
much code.

Along with the move the remaining tabulation code to use Intent.
This makes the generated code (ainly the implementation) looks a lot
better.

### DIFF
--- a/objcgen/objcgenerator.cs
+++ b/objcgen/objcgenerator.cs
@@ -54,9 +54,13 @@ namespace ObjC {
 
 			implementation.WriteLine ("static void __initialize_mono ()");
 			implementation.WriteLine ("{");
-			implementation.WriteLine ("\tif (__mono_context.domain)");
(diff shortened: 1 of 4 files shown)


objcgen/objcgenerator.cs
```diff
@@ -692,7 +653,6 @@ void GenerateDefaultValuesWrapper (string name, MethodBase mb, ParameterInfo[] p
 			headers.WriteLine ();
 
 			builder.BeginImplementation ();
-			implementation.Write ('\t');
```
dalexsoto: I guess since this wasn't replaced by any additional call it is no longer needed

spouliot: yup, that local, implied state was not always correct

PR #191, opened 2017-04-26
[objc] Add basic interface/protocol support

spouliot:
This allow an .NET API to return an interface. That interface is defined
as a protocol in ObjC.

We also supply an internal `__*Wrapper` class so ObjC can call the
interface members.

Also fix virtual calls - the current code was calling the interface
itself, not the implementation for it.

Note: it's basic in the sense that it does not allow .NET to call back
into ObjC code (that conforms to the protocol).

### DIFF
--- a/objcgen/methodhelper.cs
+++ b/objcgen/methodhelper.cs
@@ -22,6 +22,7 @@ namespace ObjC {
 		public bool IsExtension { get; set; }
 		public bool IsStatic { get; set; }
 		public bool IsValueType { get; set; }
+		public bool IsVirtual { get; set; }
 
 		public string ReturnType { get; set; }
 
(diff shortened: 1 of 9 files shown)


objcgen/methodhelper.cs
```diff
@@ -86,9 +87,14 @@ public void WriteInvoke (string args)
 				}
 			}
 
+			var method = "__method";
+			if (IsVirtual) {
+				implementation.WriteLine ($"MonoMethod* __virtual_method = mono_object_get_virtual_method ({instance}, __method);");
```
spouliot: ^ @tritao I did not see any call to this function in `master` so you likely have a similar issue with virtual calls.

tritao: Yeah I'll need to add this to the C generator too.

spouliot:
@dalexsoto pick something harder ;-)

PR #199, opened 2017-04-27
[objc] Add support for NSObject subclasses by reusing the output from the static registrar.

rolfbjarne:


objcgen/driver.cs
```diff
@@ -207,11 +207,44 @@ public void SetCompilationTarget (string value)
...
+			get {
+				switch (Platform) {
+				case Platform.iOS:
+					return "/Library/Frameworks/Xamarin.iOS.framework/Versions/Current/lib/mono/Xamarin.iOS";
+				case Platform.tvOS:
+					return "/Library/Frameworks/Xamarin.iOS.framework/Versions/Current/lib/mono/Xamarin.TVOS";
+				case Platform.watchOS:
+					return "/Library/Frameworks/Xamarin.iOS.framework/Versions/Current/lib/mono/Xamarin.WatchOS";
+				case Platform.macOS:
+					throw new NotImplementedException ("mac platform sdk"); // need to know full/mobile
```
spouliot: Should we define extra `Platform` ? and have `macOS` defaults to one of them ? or a new (mac-only right now) `--profile` argument ?

rolfbjarne: I have extra `Platform` values coming up soon :smile:

screenshot 2017-04-27 15 22 48

spouliot: @rolfbjarne you can PR the `Platform` update separately (and merge this) if you wish.

rolfbjarne: Yeah, the `Platform` update comes with XM support as well, so I think it's better in a new PR.

PR #213, opened 2017-04-28
[docs] Update ObjC documentation so it's visible from GitHub Pages

spouliot:


spouliot:
@tritao I assume you're aware of those build failures on `master` ?

Note: I'm merging since it's unrelated and @migueldeicaza wants the doc update published.

PR #216, opened 2017-04-30
[objc] Correctly expose types that conforms to (bound) protocols

spouliot:
and add unit tests. This required a few other changes like:

* Getting the managed instance, if it exists, from a native instance is
  moved to the support code (so it can be shared for multiple usages);

* Most (unless static) bound types now expose `xamarinGetGCHandle`, i.e.
  not just the one where `isEqual:` is implemented);

* `xamarinGetGCHandle` exists only inside the implementation (.m), not
  in the header (.h) file and is commented to be for internal uses only;

* EqualsHelper is modified to reuse the shared code;

### DIFF
--- a/support/objc-support.m
+++ b/support/objc-support.m
@@ -49,3 +49,15 @@ NSComparisonResult mono_embeddinator_compare_to (MonoEmbedObject *object, MonoMe
 	void* __unbox = mono_object_unbox (__result);
 	return (NSComparisonResult) *((int*)__unbox);
 }
+
+MonoObject* mono_embeddinator_get_object (id native, bool assertOnFailure)
+{
+	if (![native respondsToSelector:@selector (xamarinGetGCHandle)]) {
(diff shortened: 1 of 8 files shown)


support/objc-support.m
```diff
@@ -49,3 +49,15 @@ NSComparisonResult mono_embeddinator_compare_to (MonoEmbedObject *object, MonoMe
 	void* __unbox = mono_object_unbox (__result);
 	return (NSComparisonResult) *((int*)__unbox);
 }
+
+MonoObject* mono_embeddinator_get_object (id native, bool assertOnFailure)
+{
+	if (![native respondsToSelector:@selector (xamarinGetGCHandle)]) {
```
rolfbjarne: Should this assert if `native=nil` and `assertOnFailure=true` (I believe it currently does, but IMHO it might make more sense to return `nil` in that case)?

Issue #175, opened 2017-04-24
[objc] `IFormatProvider` helper / proposal


spouliot:
The use of `IFormatProvider` [1] is very common in some libraries. It's also something not easy to bind correctly (i.e. make it usable from ObjC) and yet easy to workaround (`null` is generally accepted) without dropping (not generating) the API. The later (not providing an API) can leave the developer without any usable API to achieve a task (e.g. lack of an overload without a `IFormatProvider`).

The idea is to:
1. detect `IFormatProvider` in API;
2. if used, then generate a `NETIFormatProvider` **protocol**;
3. use the protocol in the API that expose `IFormatProvider`

Since `GetFormat` returns a `System.Object` we cannot generate any member, so the only useful thing to supply would be `null`. 

However someone could try to give any `NSObject` instance that conforms to `NETIFormatProvider` , that can be avoided in two ways (and ideally both):

1. generated documentation, stating that `NETIFormatProvider` cannot be used for anything else but `nil`; and
2. some protection, in generated code, to ensure only `nil` is used - whatever the argument given;

```
/** NETIFormatProvider is a special protocol that is not meant to be used.
 *  It simply allows binding API that use a `IFormatProvider` argument that accepts `nil`.
 */
@protocol NETIFormatProvider <NSObject>
@end
```

[1] The same pattern can likely be reused for other types or interfaces.
[2] https://msdn.microsoft.com/en-us/library/system.iformatprovider(v=vs.110).aspx
[3] https://msdn.microsoft.com/en-us/library/system.iformatprovider.getformat(v=vs.110).aspx

rolfbjarne:
If we generate a binding that only allows `nil` for a parameter, we can remove that parameter from the bound signature, and automatically pass it ourselves.

spouliot:
Removing the parameter could cause dupes (but we could cope for them) but it's not the main reason I'd like to keep it there :)

There's the possibility of having an API that returns (from managed) a `IFormatProvider` and it should be usable in that case. E.g. UnitsNet (in our sample) has something similar, i.e. it provides an API that returns a `CultureInfo` (which implements `IFormatProvider`).

We could also provide some helpers like a `NSLocale` category for .NET interop, which would map the closest .NET `CultureInfo`. In theory it would be mean that a CA-FR locale would produce `"1,2"` for `float d = 1.2f`.

spouliot:
pr #284 was merged

PR #220, opened 2017-05-01
[objc] Don't generate _all_ types for every assembly processed

spouliot:


spouliot:
@chamons @dalexsoto PR is about setting tests for negative test results. There will also be another PR to expose types from mscorlib (ref #175 and #176) that test the change (its how I ran into the issue).

PR #227, opened 2017-05-01
[objc] Attach pointer char '*' to variable for properties

VincentDondain:
When coding, people attach the pointer char to either the type or the variable but the convention is to attach it to the variable.
See first example in https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingIvarsAndTypes.html#//apple_ref/doc/uid/20001284-BAJGIIJE
Also all properties in the iOS framework are following this convention.

You never see `NSString * myName`.

### DIFF
--- a/objcgen/objcgenerator.cs
+++ b/objcgen/objcgenerator.cs
@@ -544,7 +544,11 @@ namespace ObjC {
 			var property_type = NameGenerator.GetTypeName (pt);
 			if (types.Contains (pt))
 				property_type += " *";
-			headers.WriteLine ($") {property_type} {name};");
+
+			var spacing = " ";
+			if (property_type.EndsWith ("*", StringComparison.Ordinal))
+				spacing = string.Empty;
+			headers.WriteLine ($") {property_type}{spacing}{name};");
 
 			ImplementMethod (getter, name, false, pi);
 			if (setter == null)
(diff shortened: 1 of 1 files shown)


objcgen/objcgenerator.cs
```diff
@@ -544,7 +544,11 @@ protected override void Generate (ProcessedProperty property)
 			var property_type = NameGenerator.GetTypeName (pt);
 			if (types.Contains (pt))
 				property_type += " *";
-			headers.WriteLine ($") {property_type} {name};");
+
+			var spacing = " ";
+			if (property_type.EndsWith ("*", StringComparison.Ordinal))
```
spouliot: Since you're using `Ordinal` you can simplify this with
`property_type [property_type.Length - 1] == '*'`

or

```
var separator = property_type [property_type.Length - 1] == '*' ? String.Empty : " ";
```

There is sadly no overload with `char` for `[Starts|Ends]With` :(

objcgen/objcgenerator.cs
```diff
@@ -573,7 +577,11 @@ protected void Generate (ProcessedFieldInfo field)
 				field_type += " *";
 
 			var name = fi.Name.CamelCase ();
-			headers.WriteLine ($") {field_type} {name};");
+
+			var spacing = " ";
+			if (field_type.EndsWith ("*", StringComparison.Ordinal))
+				spacing = string.Empty;
+			headers.WriteLine ($") {field_type}{spacing}{name};");
```
spouliot: same

spouliot:
Build error fixed by PR 231

spouliot:
build

PR #246, opened 2017-05-02
Fix some language/grammar issues in the docs

timrisi:


docs/BestPracticesObjC.md
```diff
@@ -70,7 +70,7 @@ namespace Xamarin.Xml.Configuration {
 would be used like:
 
 ```
-id reader = [[Xamarin_Xml_Configuration_.Reader alloc] init];
+id reader = [[Xamarin_Xml_Configuration.Reader alloc] init];
```
spouliot: it should be `id reader = [[Xamarin_Xml_Configuration_Reader alloc] init];` (no dot, but an underscore)

PR #251, opened 2017-05-03
[objc] Allow the inclusion/generation of BCL types to help bind user code

spouliot:
As mentioned in issue #176 [1] it's something better to bring an existing
type from the BCL (like `TimeSpan`) into the generated code than try to
convert all it's uses.

The same hack can be used to provide API even if the type itself cannot
be supported, e.g. `IFormatProvider` [2]

[1] https://github.com/mono/Embeddinator-4000/issues/176
[2] https://github.com/mono/Embeddinator-4000/issues/175

### DIFF
--- a/objcgen/NameGenerator.cs
+++ b/objcgen/NameGenerator.cs
@@ -110,7 +110,7 @@ namespace ObjC {
 					case "Void":
 						return "void";
 					default:
-						return "object";
+						return t.IsInterface ? t.FullName : "object";
 					}
 				default:
 					return t.FullName;
--- a/objcgen/objcgenerator-processor.cs
+++ b/objcgen/objcgenerator-processor.cs
@@ -13,6 +13,7 @@ namespace ObjC {
 
 		List<Exception> delayed = new List<Exception> ();
 		HashSet<Type> unsupported = new HashSet<Type> ();
+		List<Type> extra_types = new List<Type> ();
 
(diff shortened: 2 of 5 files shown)


spouliot:
02:19:50 	Run test suite Tests encountered an error (The operation couldn’t be completed. (DTXProxyChannel error 1.) If you believe this error represents a bug, please attach the log file at /Users/builder/Library/Developer/Xcode/DerivedData/libmanaged-ios-fksczggmteeilcgdhxagzfykunxm/Logs/Test/0B7751B1-F9EE-4C5F-A8DC-3C41A43C3612/Session-libmanaged-iosTests-2017-05-02_221945-EYG4aK.log)

Did not happen locally. Giving it a 2nd chance, if it works I'll file it as a random issue

spouliot:
build

spouliot:
Test fails on iOS because the (managed) linker is always enabled, `--linksdk`, which makes sense (even for the simulator). This means one simply does not call any API from mscorlib.

With a bit more work we can create an XML file (for things we want to keep) and feed it to `mtouch`.

However running the linker also means the the metadata token (for the SDK at least) will be modified so those API cannot be generated with `TOKENLOOKUP`. Again that's something we can tweak when building for platforms where the managed linker is available.

Still until those are fixed this PR cannot be merged.

spouliot:
PR #327 now cover this (and is linker friendly)

PR #264, opened 2017-05-04
[objc] Clarify (with comment) that the `_object` field (even if public) is off limit

spouliot:


spouliot:
@dalexsoto that was the message for another PR :)

PR #270, opened 2017-05-05
Split the processor from the generator

spouliot:


spouliot:
merged manually due to conflicts

Issue #277, opened 2017-05-05
[Package] Add Library/Frameworks/Xamarin.Embeddinator-4000.framework/Commands to $PATH


VincentDondain:
See conversation in https://bugzilla.xamarin.com/show_bug.cgi?id=55930

Also, one internal fellow had “issues” after installing embeddinator thinking the install didn’t work because we don’t modify `$PATH`.

dalexsoto:
➕ 1 got the same comment from another internal tester

VincentDondain:
Fixed in https://github.com/mono/Embeddinator-4000/pull/280

PR #280, opened 2017-05-05
[package/installer] Touch etc/paths.d/embeddinator-4000-commands

VincentDondain:
This embeddinator-4000-commands file has a path to the e4k command folder
which includes the objcgen command.

Fixes issue 277: [Package] Add Library/Frameworks/Xamarin.Embeddinator-4000.framework/Commands to $PATH
(https://github.com/mono/Embeddinator-4000/issues/277)

### DIFF
--- a/docs/getting-started-objc-ios.md
+++ b/docs/getting-started-objc-ios.md
@@ -52,7 +52,7 @@ Run the embeddinator to create a native framework for the managed assembly:
 
 ```shell
 cd ~/Projects/hello-from-csharp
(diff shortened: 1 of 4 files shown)


spouliot:
@rolfbjarne agreed :) my comment was related to the documentation, not the installer/feature. We don't want to tell people to use the full path (when not needed) but they need to know that new shells must be opened.

VincentDondain:
@spouliot doc updated.

PR #288, opened 2017-05-08
[objc] Update command line options for compilation target

VincentDondain:
- drop all single arguments for compilation target (--dll, --shared, --static).
- use the right value names for --target.


spouliot:
@VincentDondain tests needs to be updated too :)

PR #299, opened 2017-05-09
Add support for IEquatable. Fixes #219

### LINKED ISSUE #219: [objc] Add support for `IEquatable<T>`
We have _cute_ support for `NSObject` protocol which means the managed `bool Equals (object)` generates the expected:

```
- (bool)isEqual:(id _Nullable)other;
```

However other equality methods, like the one required to implement `IEquatable<T>` [1], are still using the _default_ generated code, which looks like:

```
- (bool)equalsObj:(TimeSpan *)obj;
```

That should be changed to generate something like:

```
- (bool)isEqualToTimeSpan:(TimeSpan * _Nullable)other;
```

[1] https://msdn.microsoft.com/en-us/library/ms131187(v=vs.110).aspx

timrisi:


objcgen/EqualsHelper.cs
```diff
@@ -4,10 +4,10 @@
 
 namespace ObjC {
 	public class EqualsHelper : MethodHelper{
+		public IKVM.Reflection.Type ParameterType { get; set; }
```
spouliot: you should add ` using Type = IKVM.Reflection.Type;` at the top, or this will end up in many places in the file.

objcgen/objcgenerator.cs
```diff
@@ -439,13 +439,34 @@ protected override void Generate (ProcessedType type)
 					ObjCTypeName = managed_name,
 					ManagedTypeName = t.FullName,
 					MonoSignature = $"CompareTo({NameGenerator.GetMonoName (pt)})",
+
```
spouliot: white noise

objcgen/objcgenerator.cs
```diff
@@ -439,13 +439,34 @@ protected override void Generate (ProcessedType type)
...
+					ManagedTypeName = t.FullName,
+				};
+
+				builder.WriteHeaders ();
+				builder.WriteImplementation ();
+			}
+
 			if (equals.TryGetValue (t, out m)) {
 				var builder = new EqualsHelper (headers, implementation) {
+					ObjCSignature = $"isEqual:(id _Nullable)other",
```
spouliot: no need for `$` here

objcgen/objcprocessor.cs
```diff
@@ -143,6 +144,13 @@ protected IEnumerable<MethodInfo> GetMethods (Type t)
 					continue;
 				}
 
+				if (implement_system_iequatable_t) {
+					if (mi.Match ("System.Boolean", "Equals", new string [] { null }) && !mi.Match ("System.Boolean", "Equals", "System.Object")) {
```
spouliot: you can avoid the 2nd `Match` if you move this into an `else` for the `Equals` check (next block)

timrisi: Actually, shouldn't even need it as an Else, as long as it's checked after the Equals (since if it hits the if block for equals, it then calls 'continue'

PR #308, opened 2017-05-10
[objc] Add an assembly load hook, and use it to find assemblies in bundles.

rolfbjarne:
Remove the assembly search hook that returns the path of an assembly, and
instead have an assembly load hook, which returns an already loaded
MonoAssembly *.

This makes writing the hook a bit simpler, because string-returning hooks need
specific memory management (a hook that returns a `const char *` won't be able
to use dynamically allocated strings easily, and returning a `char *` runs
into questions about how to free that string). It's much simpler to just
return a MonoAssembly*, which mono must free when unloaded.


spouliot:
@tritao any problem switching to the new hook on master ?

tritao:
Should be fine, thanks for the heads up.

Issue #194, opened 2017-04-26
[objc] Use word "Value" in method's names after a type


VincentDondain:
Following https://github.com/mono/Embeddinator-4000/issues/193 it's pretty common/convention in ObjC to say `Value` after a type if it's what you **have to** use in your method's name.

So https://github.com/mono/Embeddinator-4000/blob/objc/tests/managed/methods.cs#L93-L94 for instance should become:

```
- (int)doItWithInt32Value:(int) i;
- (int)doItWithStringValue:(NSString *) s;
```

spouliot:
Q: When does it make sense ? and when does it not ?

E.g. why not for the `NSString` overload ?

IMO `doItWithNSStringValue` would be quite bad, i.e. the `NS` should not be there

but `doItWithStringValue` ? there's quite a few Apple API with similar signature

Still I prefer `doItWithString` - and there's also many similar signature (without `Value`) in Apple's API

VincentDondain:
We would do it for the `NSString` overload, I fixed the description, was a typo sorry.

VincentDondain:
The rule would be to use `Value` when we're using the type's name as method name. If we're using the argument's name we don't need to add `Value`.

VincentDondain:
You're not using type's names in ObjC methods very often (when you're not generating code :P) and I believe that when you do see a type name in the method's name you expect to see `Value` next to it.

You can disagree with me and close this issue though, I'm just reading the naming methods document but if Apple also break that rule most of the time... (:

spouliot:
No disagreement just trying to extract rules that we can codify :-)

> was a typo sorry.

It's an interesting typo because it (a) looks nice and (b) commonly used by Apple. I mainly saw it because it was near another (non string) case, otherwise I think I would have fell under my radar.

Anyway it applies mostly (but not uniquely) to `NSString` and it's simpler (and consistent, easier when generating code) to have a single rule.

> You're not using type's names in ObjC methods very often

Yeah, we work with what we have :) Right now priority is given to parameter names but those can produce duplicates and type names guarantee an unique signature in .NET.

`bool Compare (int value);` and
`bool Compare (string value);`

VincentDondain:
@spouliot I believe we could use the same input types as https://github.com/mono/Embeddinator-4000/blob/objc/objcgen/NameGenerator.cs#L10-L31 (used for https://github.com/mono/Embeddinator-4000/issues/195).

We'd have some mapping like this:

```
[...]

PR #315, opened 2017-05-11
[objc] Use word 'Value' after type name in methods name

VincentDondain:
Fixes issue 194: [objc] Use word "Value" in method's names after a type
(https://github.com/mono/Embeddinator-4000/issues/194)

- We now have nice names when using types in methods names (no more Int16, Int32, Int64...)
  and they use the word 'Value' at the end which is standard in Objective-C.
  See: https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingMethods.html#//apple_ref/doc/uid/20001282-BCIGIJJF
  The method's argument description also use the new friendly names which is better since it matches the Objective-C types and the argument's names.
- Use camel case for method's argument description so we have `findFirstName: lastName:`.
- Fixed up tests accordingly.

### DIFF
--- a/objcgen/NameGenerator.cs
+++ b/objcgen/NameGenerator.cs
@@ -30,6 +30,27 @@ namespace ObjC {
 			{ "signed char", "aChar" }
 		};
 
+		public static Dictionary<string, string> ObjCTypeToMethodName = new Dictionary<string, string> {
+			{ "int", "IntValue" },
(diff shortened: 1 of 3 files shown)


dalexsoto:
Oh, the web editor to resolve conflicts is actually pretty good! 😮

PR #316, opened 2017-05-11
AddExtraType was trying to queue types to typesQueue before it was initialized

timrisi:
…eated

### DIFF
--- a/objcgen/processor.cs
+++ b/objcgen/processor.cs
@@ -67,7 +67,7 @@ namespace Embeddinator {
 
 		protected void AddExtraType (ProcessedType pt)
 		{
-			typeQueue.Enqueue (pt);
+			Types.Add (pt);
 			// extra types are (most likely) outside the input list of assemblies
 			AddIfUnique (pt.Assembly);
 		}
--- a/objcgen/processor.cs
+++ b/objcgen/processor.cs
@@ -1,4 +1,4 @@
-﻿using System;
+﻿﻿using System;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


objcgen/processor.cs
```diff
@@ -67,7 +67,7 @@ public void Process (ProcessedAssembly a)
 
 		protected void AddExtraType (ProcessedType pt)
 		{
-			typeQueue.Enqueue (pt);
+			Types.Add (pt);
```
spouliot: That does not work, e.g. when `AddExtraType` is called while processing the `typeQueue`. `Types` was cleared and won't be used again while dequeuing.

I _think_ what's needed to create an empty queue and add `Types` to it (then clear it). What was the test case ? just run on mscorlib.dll ?

timrisi: Do you mean create a queue in the AddExtraType method here and add types to it (and clear it)?  That would cause issues when it later (if called while processing an assembly) calls ```typeQueue = new Queue<ProcessedType> (Types);```.

I don't see an easy way to add the contents of Types to an existing queue without having to iterate over Types and add them one at a time.

What about a null check for typeQueue and if it's null add to Types (should catch when processing assembly before the type is created), otherwise add it to the queue?

spouliot: No, I mean always create the `typeQueue` (like `assemblyQueue`). That will make `AddExtraType` work. It also requires to enqueue all items in `Types` before clearing it.

Ideally `AddExtraType` should always be callable, since there can more than than one (current) reason to use it.

PR #319, opened 2017-05-11
[objc] Make the processor linker-aware

spouliot:
Processed assemblies and types that are part of the SDK (i.e. not user
code) needs to preserve the types / members that are bound.

This can be done automatically by the processor as those types/members
are being added (before generation). At the end of the process we can
use the collected data to generate a `bindings.xml` file that the linker
from Xamarin.iOS or Xamarin.Mac can consume.

Notes:
* Neither ObjC processing or generating code is unaffected by this;
* Unit tests to be included with TimeSpan support (separate PR)


spouliot:
Part 1: this should be merged after PR #292 (decimal support) so it can replace the static `linker.xml` file.

Part 2: will add `System.TimeSpan` support and, unlike PR #251 (first attempt), it will work correctly for iOS.

spouliot:
PR #325 will fix the (unrelated) failure above.

PR #324, opened 2017-05-11
[objc] Add support for creating Xamarin.Mac frameworks.

rolfbjarne:
* Add multiple new platforms, one for each Xamarin.Mac flavor
  (modern/full/system), and make these new platform values build a framework
  using mmp.

* Add support for creating a framework for the existing `macOS` platform
  (which does not include Xamarin.Mac).

* Rework testing significantly, adding unit tests for Xamarin.Mac framework support:

    * Make managed.csproj a shared library project, and make another project
      managed-*.csproj for each platform we support that references this
      shared project.

    * Add managed unit tests that runs the embeddinator on the corresponding
      managed-*.csproj, generates an Xcode test project, and runs the tests in
      the generated Xcode project. This is a replacement for a few identical
      tests in tests/objc-cli/Makefile (and those are now removed).


objcgen/driver.cs
```diff
@@ -152,6 +162,18 @@ public void SetPlatform (string platform)
...
+				Platform = Platform.macOSModern;
+				break;
+			case "macosfull":
+			case "macos-full":
+				Platform = Platform.macOSFull;
+				break;
+			case "macossystem":
+			case "macos-system":
+				Platform = Platform.macOSSystem;
+				break;
```
spouliot: This needs to be included in our `/docs/` documentation. A large part can likely just link to back the current XM documentation (for the differences between XM profiles). What's need to be clear is the difference/when XM is required.

Maybe a `Targets.md` document that list the choices and describe them ?

rolfbjarne: I've added some documentation.

PR #332, opened 2017-05-11
[objc] Add decimal array ref support

dalexsoto:


objcgen/objcgenerator.cs
```diff
@@ -667,7 +671,7 @@ void GenerateArgumentArray (string parameterName, string argumentName, Type t, b
 				implementation.WriteLine ($"mono_array_set ({pnameArr}, {typeName}, {pnameIdx}, {pnameRet}.{returnValue});");
 				break;
 			case TypeCode.Decimal:
-				implementation.WriteLine ($"NSDecimalNumber* {pnameRet} = {parameterName}[{pnameIdx}];");
+				implementation.WriteLine ($"NSDecimalNumber* {pnameRet} = {(is_by_ref ? $"(*{parameterName})" : parameterName)}[{pnameIdx}];");
```
spouliot: minor/style: not sure I like using a `$` inside another one `$`, makes it harder to read than it needs to be. Either:

* add a line before to compute it; or
* always keep the parenthesis and make only `*` conditional on `is_by_ref`

dalexsoto:
@spouliot fixed

PR #344, opened 2017-05-15
[objc] Add NSDate <-> DateTime support.

dalexsoto:
# DateTime v.s. NSDate

From [NSDate's](https://developer.apple.com/reference/foundation/nsdate?language=objc) documentation:

> NSDate objects encapsulate a single point in time, independent of any particular calendrical system or time zone. Date objects are immutable, representing an invariant time interval relative to an absolute reference date (00:00:00 UTC on 1 January 2001).

Due to NSDate reference date, all conversions between it and DateTime must be done in UTC.

## DateTime to NSDate

When converting from `DateTime` to `NSDate` the DateTime's `Kind` property is taken into account.

| Kind         | Results                                                                                            |
|------------: | -------------------------------------------------------------------------------------------------- |
| Utc          | Conversion is performed using the provided DateTime object as is.                                  |
| Local        | The result of calling `ToUniversalTime ()` in the provided DateTime object is used for conversion. |
| Unspecified  | The provided DateTime object is assumed to be UTC, so same behavior as Kind == Utc.                |

The conversion is done by using the following formula:

**TimeInterval** = DateTimeObjectTicks - NSDateReferenceDateTicks[dt] / [TicksPerSecond](https://msdn.microsoft.com/en-us/library/system.timespan.tickspersecond(v=vs.110).aspx)

Once we have the TimeInterval we use NSDate's [dateWithTimeIntervalSinceReferenceDate:](https://developer.apple.com/reference/foundation/nsdate/1591577-datewithtimeintervalsincereferen?language=objc) selector to create it.

## NSDate to DateTime

Going from NSDate to DateTime we assume we are getting a NSDate instance which's reference date is **00:00:00 UTC on 1 January 2001** and use the following formula:

**DateTimeTicks** = NSDateTimeIntervalSinceReferenceDate * [TicksPerSecond](https://msdn.microsoft.com/en-us/library/system.timespan.tickspersecond(v=vs.110).aspx) + NSDateReferenceDateTicks[dt]

Once we calculated the **DateTimeTicks** we use the following DateTime [constructor](https://msdn.microsoft.com/en-us/library/w0d47c9c(v=vs.110).aspx) setting its `kind` to `DateTimeKind.Utc`.

There are some considerations that you must be aware of, NSDate can be `nil` but a DateTime is a struct in .NET and by definition it can't be `null`. If you give a `nil` NSDate we will translate it to the default DateTime value which maps to `DateTime.MinValue`.

MinValue and MaxValue are also different, NSDate can support greater and lower boundaries than DateTime's so whenever you give a greater or lower value we will set it to DateTime's [MaxValue](https://msdn.microsoft.com/en-us/library/system.datetime.maxvalue(v=vs.110).aspx) or [MinValue](https://msdn.microsoft.com/en-us/library/system.datetime.minvalue(v=vs.110).aspx) respectively.

**dt**: NSDate reference date **00:00:00 UTC on 1 January 2001** => `new DateTime (year:2001, month:1, day:1, hour:0, minute:0, second:0, kind:DateTimeKind.Utc).Ticks;`


support/mono-support.h
```diff
@@ -61,6 +61,10 @@ MonoClass*      mono_get_uint64_class (void);
 MonoClass*      mono_get_single_class (void);
 MonoClass*      mono_get_double_class (void);
 int             mono_array_element_size (MonoClass *ac);
+void			mono_field_static_get_value (MonoVTable *vt, MonoClassField *field, void *value);
+void			mono_runtime_class_init (MonoVTable *vtable);
+int				mono_class_instance_size (MonoClass *klass);
+int				mono_class_array_element_size (MonoClass *klass);
```
rolfbjarne: There seems to be some space/tab issues here.

support/mono_embeddinator.c
```diff
@@ -314,3 +314,67 @@ MonoClass* mono_embeddinator_get_decimal_class ()
...
+		if (dtmaxtickserr)
+			mono_embeddinator_throw_exception (dtmaxtickserr);
+
+		MonoObject* dtmintickserr = nil;
+		MonoObject* dtminticksboxed = mono_runtime_invoke (dtticks, dtminvalueptr, NULL, &dtmintickserr);
+		if (dtmintickserr)
+			mono_embeddinator_throw_exception (dtmintickserr);
+
+		datetimedefaults.MaxValueTicks = *(long long *) mono_object_unbox (dtmaxticksboxed);
+		datetimedefaults.MinValueTicks = *(long long *) mono_object_unbox (dtminticksboxed);
```
rolfbjarne: From what I can tell, all these values are constant and will never change.

This means you can just hard code them:

```c
datetimedefaults.NetTicksPerSecond = 10000000;
datetimedefaults.MaxValueTicks = 3155378975999999999;
datetimedefaults.MinValueTicks = 0;
```

from:

```shell
$ csharp -e 'DateTime.MaxValue.Ticks'
3155378975999999999
$ csharp -e 'DateTime.MinValue.Ticks'
0
$ csharp -e 'TimeSpan.TicksPerSecond'
10000000
```

support/objc-support.m
```diff
@@ -146,3 +146,129 @@ MonoDecimal mono_embeddinator_get_system_decimal (NSDecimalNumber* nsdecimalnumb
...
+		mono_embeddinator_throw_exception (ex);
+
+	MonoObject* ex2 = nil;
+	MonoObject* boxedticks = mono_runtime_invoke (datetimeticks, datetime, NULL, &ex2);
+	if (ex2)
+		mono_embeddinator_throw_exception (ex);
+
+	long long ticks = *(long long *) mono_object_unbox (boxedticks);
+
+	return ticks;
```
rolfbjarne: This value can also be hardcoded.

support/objc-support.m
```diff
@@ -146,3 +146,129 @@ MonoDecimal mono_embeddinator_get_system_decimal (NSDecimalNumber* nsdecimalnumb
...
+}
+
+static long long nsdaterefdateticks = 0;
+
+NSDate* mono_embeddinator_get_nsdate (void* datetime, mono_embeddinator_context_t* context)
```
rolfbjarne: Can you add a strongly defined type for `datetime` here (and everywhere else you're using `void*`) instead of `void*`? That makes errors harder (and more visible in the code review), and it's also easier to debug code (in lldb).

support/objc-support.m
```diff
@@ -146,3 +146,129 @@ MonoDecimal mono_embeddinator_get_system_decimal (NSDecimalNumber* nsdecimalnumb
...
+	if (!dtkindmethod)
+		dtkindmethod = mono_embeddinator_lookup_method ("System.DateTimeKind:get_Kind()", datetimeclass);
+
+	MonoObject* kindex = nil;
+	MonoObject* kindboxed = mono_runtime_invoke (dtkindmethod, datetime, NULL, &kindex);
+	if (kindex)
+		mono_embeddinator_throw_exception (kindex);
+	E4KDateTimeKind kind = *(E4KDateTimeKind *) mono_object_unbox (kindboxed);
+
+	if (kind == E4KDateTimeKind_Local) {
```
rolfbjarne: The number of ticks doesn't change depending on the kind:

```shell
> csharp
Mono C# Shell, type "help;" for help

Enter statements below.
csharp> var dt = DateTime.Now;
csharp> dt.Ticks;
636305384874966380              
csharp> dt.ToUniversalTime ().Ticks;
636305312874966380                                 
csharp> dt.ToLocalTime ().Ticks;                   
636305384874966380                                 
csharp>                                            
```

The kind of a DateTime only affects how that value is rendered, not the underlying numerical representation (except the `kind` field of course).

support/objc-support.m
```diff
@@ -146,3 +146,129 @@ MonoDecimal mono_embeddinator_get_system_decimal (NSDecimalNumber* nsdecimalnumb
...
+	void* datetimeargs [2];
+	if (!nsdate)
+		datetimeargs [0] = &datetimeDefs->MinValueTicks;
+	else {
+		long long nsdateinterval = [nsdate timeIntervalSinceReferenceDate];
+		long long dateticks = nsdateinterval * datetimeDefs->NetTicksPerSecond + nsdaterefdateticks;
+
+		if (dateticks > datetimeDefs->MaxValueTicks)
+			dateticks = datetimeDefs->MaxValueTicks;
+		if (dateticks < datetimeDefs->MinValueTicks)
```
rolfbjarne: This can be an `else if`

support/objc-support.m
dalexsoto: Sneaky digits 😉 It does change.

`csharp> dt.Ticks;`
6363053**848**74966380              
`csharp> dt.ToUniversalTime ().Ticks;`
6363053**128**74966380                                 
`csharp> dt.ToLocalTime ().Ticks;`                 
6363053**848**74966380

support/mono_embeddinator.c
dalexsoto: Yeah, I was trying to play safe by actually get the values from the runtime but I guess I was just over-engeneering this so if no objections I will harcode the values.

support/objc-support.m
dalexsoto: Yeah, I was trying to play safe by actually get the values from the runtime but I guess I was just over-engeneering this so if no objections I will harcode the values.

support/mono_embeddinator.c
spouliot: you might even want to use `#define` on them and kill the structure

tests/objc-cli/libmanaged/Tests/Tests.m
```diff
@@ -505,6 +505,47 @@ - (void)testTypes {
...
 	[Type_Decimal reverseDecimalArrRefDecArr:&decimalrefNilarr];
 	XCTAssertNil(decimalrefNilarr, "decimalrefNilarr");
+
+	NSDate *distantFuture = [Type_DateTime returnDateDatetime:[NSDate distantFuture]];
+	XCTAssertNotNil(distantFuture, "distantFuture");
+
+	NSDate *verydistantFuture = [NSDate dateWithTimeIntervalSinceReferenceDate:[distantFuture timeIntervalSinceReferenceDate] * 8];
+	NSDate *netmaxverydistantFuture = [Type_DateTime returnDateDatetime:verydistantFuture];
+	NSDate *nsnetmax = [NSDate dateWithTimeIntervalSinceReferenceDate:(NSTimeInterval) 252423993600];
+	XCTAssertEqualWithAccuracy([nsnetmax timeIntervalSinceReferenceDate], [netmaxverydistantFuture timeIntervalSinceReferenceDate], 0.001, "netmaxverydistantFuture");
```
spouliot: minor/style: space before `(`
here and later in this file

docs/ObjC.md
```diff
@@ -180,3 +180,41 @@ becomes:
...
+#### NSDate to DateTime
+
+Going from NSDate to DateTime we assume we are getting a NSDate instance which's reference date is **00:00:00 UTC on 1 January 2001** and use the following formula:
+
+**DateTimeTicks** = NSDateTimeIntervalSinceReferenceDate * [TicksPerSecond](https://msdn.microsoft.com/en-us/library/system.timespan.tickspersecond(v=vs.110).aspx) + NSDateReferenceDateTicks[dt]
+
+Once we calculated the **DateTimeTicks** we use the following DateTime [constructor](https://msdn.microsoft.com/en-us/library/w0d47c9c(v=vs.110).aspx) setting its `kind` to `DateTimeKind.Utc`.
+
+There are some considerations that you must be aware of, NSDate can be `nil` but a DateTime is a struct in .NET and by definition it can't be `null`. If you give a `nil` NSDate we will translate it to the default DateTime value which maps to `DateTime.MinValue`.
+
```
dalexsoto: Are you ok if I do this in a separated PR? also need to do this for `decimal`

spouliot: Fine in 2nd PR

support/objc-support.m
rolfbjarne: You're right of course 👍

support/objc-support.m
```diff
@@ -146,3 +146,91 @@ MonoDecimal mono_embeddinator_get_system_decimal (NSDecimalNumber* nsdecimalnumb
...
+		datetimeargs [0] = &dateticks;
+	}
+
+	datetimeargs [1] = &utc;
+
+	if (!datetimector)
+		datetimector = mono_embeddinator_lookup_method ("System.DateTime:.ctor(long,System.DateTimeKind)", datetimeclass);
+
+	MonoObject* datetimeshell = mono_object_new (context->domain, datetimeclass);
+	E4KDateTime datetime = *(E4KDateTime *) mono_object_unbox (datetimeshell);
```
rolfbjarne: Structures are initialized to 0 in managed code, which means you can just replicate the same in C:

```c
E4KDateTime datetime = { 0 };
```

and skip two calls into mono (both `mono_object_new` and `mono_object_unbox`).

dalexsoto:
@rolfbjarne / @spouliot Finally green :) Are you ok on merging this?

PR #364, opened 2017-05-23
CLI - support for multiple generators

jonathanpeppers:
- You can specify `--gen` multiple times at the command line
- `--gen=Java` will automatically imply `--gen=C` first
- I don't think this breaks any command line argument validation

@tritao does it need to assume C if you leave `--gen` off completely? This would print `Please specify a target generator.`, but I think that is probably right.


tritao:
Yeah that seems like a reasonable / better behavior.

PR #375, opened 2017-05-30
Java - changes for Java compilation and creating jar files

jonathanpeppers:
I am mainly opening this for feedback, and to get Java in a better place. @tritao let me know what you think.

A couple questions here:
- Is `/deps/java` a fine place to put the JNA library? Setting up a Maven build is probably the only way to not have to commit the jar file. I don't know we if we want to depend on Maven being installed.
- I used `FindDirectory` to locate `/support/java` and `/deps/java`, is there a better way to do this? This supports running `MonoEmbeddinator4000.exe` from the root of the repo and from the output dir
- Is compiling Java to `OUTPUT_DIR/classes/` fine?
- Is the jar filename `OUTPUT_DIR/classes.jar` work for Java in general? It will probably be named `classes.jar`  for `--platform=Android` because it will get packaged inside an `AAR`.


tritao:
Hey Jonathan,

This looks good, just a couple points:

* We already have the JNA available in `external/jna` so there's no need to add a new one.
* I think compiling to `classes` should be OK.
* For desktop Java I think naming the jar the same as the native library would be better than `classes.jar` which AFAIK is just an Android convention.

PR #376, opened 2017-05-31
Java - jar files are now usable within a Java project on MacOS

jonathanpeppers:
The changes here I discussed with @tritao, these are mainly to get desktop Java in a working state.

## Changes
- On MacOS, `darwin/libX.dylib` is packaged up in the jar file
- .NET assemblies are packaged into `/assemblies/` inside the jar file
- *For now*, `Runtime.java` extracts the assemblies to a temp directory for use. We should investigate if the runtime can load them directly from the jar.

You can find the sample I used to test [here](https://github.com/jonathanpeppers/Embeddinated-Android/blob/master/lib-java/src/main/java/com/xamarin/Main.java).

### DIFF
--- a/support/java/Runtime.java
+++ b/support/java/Runtime.java
@@ -29,7 +29,10 @@
 package mono.embeddinator;
 
 import com.sun.jna.*;
-import com.sun.jna.ptr.*;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
 import java.util.*;
 
 public final class Runtime {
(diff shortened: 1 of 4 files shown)


support/java/Runtime.java
```diff
@@ -96,9 +99,28 @@ protected List getFieldOrder() {
...
 
+        String tmp = System.getProperty("java.io.tmpdir");
+        String assemblyPath = Utilities.combinePath(tmp, library);
+        File assemblyFile = new File(assemblyPath + ".dll");
+
+        if (!assemblyFile.exists()) {
+            String resourcePath = "/assemblies/" + library + ".dll";
+            InputStream stream = Runtime.class.getResourceAsStream(resourcePath);
+            if (stream == null) {
+                pendingException.set(new RuntimeException("Unable to locate " + resourcePath + " within jar file!"));
```
jonathanpeppers: @tritao I was unsure if this is the best mechanism for gracefully crashing. When testing, if this exception occurred it wasn't printing a readable error message at the console.

tritao: `pendingException` is for pending managed exceptions. Here we should be able to just throw a regular Java exception.

tritao:
CI seems to be failing with:
`.Embeddinator error: Mono failed to load assembly.`

PR #380, opened 2017-06-01
[java] - embed JNA into our jar file

jonathanpeppers:
- We don’t want developers to have to pull jna-4.4.0.jar into their
project
- This unzips it into our jar, so they just have to worry about a
single file
- Updated tests, so it is not using the jar from /external/jna


tritao:
whitelist

PR #388, opened 2017-06-07
[java] improvements for Java on MacOS

jonathanpeppers:
### Mark .NET assemblies to be deleted on exit
- After a lot of research, I discovered that JNA actually extracts
native libraries from jar files into a temporary directory [see here](https://github.com/java-native-access/jna/blob/master/src/com/sun/jna/Native.java#L998)
- This makes it seem reasonable that we are doing the same with .NET
assemblies
- One improvement I saw from their implementation that could improve
ours: don’t use `File.exists()` and use `File.deleteOnExit()`
- This should extract/delete .NET assemblies on every run the same way
that JNA handles native libraries

### Package *all* .NET assemblies into jar file
- Turns out that desktop Java was loading `mscorlib.dll` from current
machine’s Mono installation
- Now we package mono assemblies into jar file
- Same logic is used for calling `mono_set_dirs` as Android

### DIFF
--- a/binder/Compilation.cs
+++ b/binder/Compilation.cs
@@ -447,6 +447,27 @@ namespace MonoEmbeddinator4000
                 {
                     File.Copy(assembly, Path.Combine(assembliesDir, Path.GetFileName(assembly)), true);
                 }
+
+                //Copy any referenced assemblies such as mscorlib.dll
+                List<string> referencedAssemblies = new List<string>();
+                foreach (var assembly in Assemblies)
+                {
+                    foreach (var reference in assembly.GetReferencedAssemblies())
+                    {
+                        if (!referencedAssemblies.Contains(reference.Name))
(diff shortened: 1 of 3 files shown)


jonathanpeppers:
@tritao can you look at the failed Travis CI build?

It almost looks like there is a `\n` in the middle of the path to `javac`:
```
[ERROR] FATAL UNHANDLED EXCEPTION: System.ComponentModel.Win32Exception (0x80004005): ApplicationName='/Library/Java/JavaVirtualMachines/jdk1.8.0_92.jdk/Contents/Home
/bin/javac', CommandLine='...'
```

We could see if it happens on #389, too.

tritao:
My bad, should be fixed with https://github.com/mono/Embeddinator-4000/commit/b69db9e85efe893852528ed205ddae0d743f578d.

Feel free to merge if CI is green.

PR #411, opened 2017-06-27
[external] Bump to latest CppSharp.

tritao:


build/projects/CppSharp.Parser.CSharp.csproj
```diff
@@ -41,15 +41,15 @@
...
+    </Compile>
     <None Include="../../external/CppSharp/src/CppParser/Bindings/CSharp/premake5.lua">
       <Link>premake5.lua</Link>
     </None>
-    <Compile Include="../../external/CppSharp/src/CppParser/Bindings/CSharp/x86_64-apple-darwin12.4.0/CppSharp.CppParser.cs">
-      <Link>x86_64-apple-darwin12.4.0/CppSharp.CppParser.cs</Link>
-    </Compile>
-    <Compile Include="../../external/CppSharp/src/CppParser/Bindings/CSharp/x86_64-apple-darwin12.4.0/Std.cs">
-      <Link>x86_64-apple-darwin12.4.0/Std.cs</Link>
-    </Compile>
```
realvictorprm: Is this correct?

build/projects/CppSharp.Parser.CSharp.csproj
```diff
@@ -41,15 +41,15 @@
   <ItemGroup>
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="../../external/CppSharp/src/CppParser/Bindings/CSharp/i686-apple-darwin12.4.0/CppSharp.CppParser.cs">
+      <Link>i686-apple-darwin12.4.0/CppSharp.CppParser.cs</Link>
+    </Compile>
+    <Compile Include="../../external/CppSharp/src/CppParser/Bindings/CSharp/i686-apple-darwin12.4.0/Std.cs">
+      <Link>i686-apple-darwin12.4.0/Std.cs</Link>
+    </Compile>
```
realvictorprm: So only x86?

build/projects/CppSharp.Parser.CSharp.csproj
tritao: Hmm good catch, guess the default architecture for this was changed in CppSharp. But Embeddinator doesn't actually rely on any of `CppSharp.Parser.CSharp` at runtime, it's only included in the build right now because its a dependency of `CppSharp.Generator`.

realvictorprm: okay @tritao
as long as the build and the tests pass :)

Issue #406, opened 2017-06-25
[java] C# decimal is not exposed


mattleibow:
When generating types that have public `decimal` members, they get skipped:

> Property 'Operand' was ignored due to ignored type

It appears that you may be able to use Java's `BigDecimal`: http://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html

migueldeicaza:
I am not sure that we should do decimal mapping here, because the conversions between C# decimal and Java bigdecimal  will not be lossless.   Perhaps we need to surface decimal as "DNDecimal".   Initially it could be very empty with nothing in there, other than "ToBigDecimal()"

PR #412, opened 2017-06-27
[c][java] Implement Decimal types support

tritao:
Partial fix for https://github.com/mono/Embeddinator-4000/issues/406


knocte:
@tritao Why is it partial only? What is not supported a.t.m.?

PR #414, opened 2017-06-29
Cake build scripts

jonathanpeppers:
Our various `Makefiles` do not currently work well on windows, so @tritao and I are looking to start the process of migrating to Cake.

This PR includes:
- Just a few build targets `Generate-Android` is set to `Default`
- `Run-Java-Tests` works (except on Windows there is an issue compiling C)
- Windows & OS X bootstrap scripts for Cake
- Point `external/Xamarin.Android.Tools` back to official repo

Changes from standard Cake scripts:
- Using experimental flag for C# 6/Roslyn on Windows
- Using `.cake` instead of the `tools` directory
- Using `.nuget/nuget.exe` because we already have it there

From here, we can start adding more Cake tasks.

### DIFF
--- a/.cake/packages.config
+++ b/.cake/packages.config
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<packages>
+    <package id="Cake" version="0.20.0" />
+</packages>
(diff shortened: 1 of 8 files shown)


realvictorprm:
nice

PR #415, opened 2017-06-30
[build] Added couple more Cake tasks for running the test suite

tritao:


realvictorprm:
@tritao give me some cake :smile: Looks nice!

jonathanpeppers:
One thing I'll mention is a bug I found with Cake running on Mono, you might have hit it here and had to workaround it.

You can't have a global var:
```csharp
var somestring = "mono!";

void SomeMethod()
{
    //And then use it in a helper method
    Console.WriteLine(somestring);
}
```
This works on windows, but in Mono it says `somestring` does not exist.

So yeah, you have to move the var inside the method, like you did with `premakepath`.

tritao:
Thanks for the heads-up. I did have some fun bugs/errors every time I had some syntax issues on the source, which at first were really puzzling me.

Issue #418, opened 2017-07-01
[Android] condition 'vtable' not met


Zomb:
The native app consuming the generated aar is crashing with (some of) the stacktrace shown below. I can't really figure out what might be causing the crash, is there any way for me to get more information on this crash?

```
> 07-01 18:48:24.922 1175-1230/? W/AppOps: Finishing op nesting under-run: uid 1000 pkg android code 24 time=0 duration=0 nesting=0
>                                          
>                                          [ 07-01 18:48:25.012  7074: 7074 F/         ]
>                                          * Assertion at /Users/builder/jenkins/workspace/xamarin-android/xamarin-android/external/mono/mono/metadata/object.c:5223, condition `vtable' not met
> 07-01 18:48:25.012 7074-7074/com.starkey.android.newlink.nativeandroid E/mono-rt: Stacktrace:
> 07-01 18:48:25.012 7074-7074/com.starkey.android.newlink.nativeandroid E/mono-rt: /proc/self/maps:
> 07-01 18:48:25.012 7074-7074/com.starkey.android.newlink.nativeandroid E/mono-rt: 12c00000-12c9f000 rw-p 00000000 00:04 18926                              /dev/ashmem/dalvik-main space (deleted)
> 07-01 18:48:25.012 7074-7074/com.starkey.android.newlink.nativeandroid E/mono-rt: 12c9f000-12e9f000 rw-p 0009f000 00:04 18926                              /dev/ashmem/dalvik-main space (deleted)
> 07-01 18:48:25.012 7074-7074/com.starkey.android.newlink.nativeandroid E/mono-rt: 12e9f000-22c00000 ---p 0029f000 00:04 18926                              /dev/ashmem/dalvik-main space (deleted)
> 07-01 18:48:25.012 7074-7074/com.starkey.android.newlink.nativeandroid E/mono-rt: 32c00000-32c01000 rw-p 00000000 00:04 18927                              /dev/ashmem/dalvik-main space 1 (deleted)
> 07-01 18:48:25.012 7074-7074/com.starkey.android.newlink.nativeandroid E/mono-rt: 32c01000-42c00000 ---p 00001000 00:04 18927                              /dev/ashmem/dalvik-main space 1 (deleted)
[...]

PR #426, opened 2017-07-11
[java] improvements to package names

jonathanpeppers:
### Removes `_dll` from Java package names

Previously if you had `hello.dll` with:
```csharp
namespace Example { 
    public class Hello { 
        public string World() { 
            return "Hello, World!"; 
        } 
    } 
}
```

Embeddinator would generate a class like this:
```java
package hello_dll.example;

public class Hello {
    public String world() {
        //Code here
    }
}
```

The `_dll` is redundant, so this PR removes it.

### Other Changes

- Refactor to use utility methods for converting to Java package names (`JavaGenerator.GetNativeLibPackageName` and `JavaNative.GetNativeLibClassName`)
- Removed the `com.` prefix for Android package names
- `<meta-data>` tag now more accurately includes the full name of the main Java class and is renamed to `mono.embeddinator.classname`
- updated tests to not use `_dll`
- more C# tests: for JavaGenerator and AndroidManifest.xml
- fixed missing CppSharp project from `generator.sln`


tests/MonoEmbeddinator4000.Tests/DriverTest.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        [Test]
+        public void UpperCase_java()
+        {
+            RunDriver("HelloUpper");
+
+            string path = Path.Combine(options.OutputDir, driver.Output.Files.Keys.First());
+            string java = File.ReadAllText(path);
+            Approvals.Verify(java); //TODO: I don't know if "String wORLD()" is what we want
```
tritao: Yeah I think there's a bug in upper case handling here, but it can be fixed separately from this PR.

Zomb:
I thought I would post this since you are working on renames, and it might not be an issue anymore, but as of yesterday, if you named a C# assembly with a "." like Android.TestProject you would get an error with e4k. I just worked around it by removing the . in the assembly name.

jonathanpeppers:
Thanks @Zomb I will add a test case for this and see what breaks.

With the work on our test suite, it should be a lot easier to test assembly names. If I remember correctly, we probably need a test with an assembly with upper/lower case, too.

jonathanpeppers:
@Zomb it worked, but these are good test cases: https://github.com/mono/Embeddinator-4000/pull/428

I wanted to check, when you had the issue did it break at the point of running Embeddinator? or did it break in Android Studio?

If it's the latter, the tests on the #428 PR might not cover it.

Zomb:
@jonathanpeppers Now that I think about it, it was in Android Studio, not e4k. Sorry about that. I was exactly like the UpperCase issue in #418 and that showed up in Android Studio too.

PR #430, opened 2017-07-11
[android] better selection of Android API level

jonathanpeppers:
- Moved API level selection to `XamarinAndroid` class
- `CompileNDK` now works off the highest API level available instead of
hardcoding 24
- added some tests


jonathanpeppers:
Yeah I think it needs to do these at some point:
- Command line option for api level
- Look at `TargetFrameworkVersion` on the assembly passed in

PR #437, opened 2017-07-13
[android] do not include PDB files unless -debug is set

jonathanpeppers:
- Best option was to delete them since the `LinkAssemblies` task is
copying them
- Added `Approvals.VerifyZipFile`
- Added tests validating the contents of JAR and AAR files
- We now have a test that checks the `-debug` switch is preserving PDB
files
- Some of these new tests are slow, so I marked with
`[Category(“Slow”)]` for use later if needed


tritao:
👍

PR #441, opened 2017-07-14
[android] seeing if AppVeyor can run android tests

jonathanpeppers:
Don't merge this, I will keep rebasing this to see if we can get Android/emulator tests running on AppVeyor.


jonathanpeppers:
I will maybe revisit this later, doesn't seem like they support it: http://help.appveyor.com/discussions/questions/5638-android-emulator

PR #444, opened 2017-07-19
[c][java] Type checking and code generation fixes

tritao:
This PR provides a couple improvements when propagating ignored types, and checking for them at generation time.

It undoes some of your changes introduced in https://github.com/mono/Embeddinator-4000/pull/442, but should be more robust going forward allowing us to ignore declarations for reasons other than unsupported types.

Also some other generation fixes for getting mscorlib.dll and System.dll to properly compile.


jonathanpeppers:
Looks good if CI passes. These failing tests are fixed.

PR #447, opened 2017-07-20
[java] fixes for C# interfaces

jonathanpeppers:
The following were generating invalid Java:
- C# interfaces extending a base interface
- C# methods that implement an interface explicitly

This broke things like:
- Many libraries have interface inheritance, especially things that extend `IDisposable`
- Xamarin.Forms has explicit interface methods on `VisualElement`, `Element`, and `BindableObject`

Changes:
- `InterfacesPass` needs to loop over any base interfaces
- `InterfacesPass` needs to include a check for `__getObject`, interface inheritance was causing this method to get generated multiple times
- Added `IsExplicitInterfaceMethod` extension method
- `AstGenerator` will now mark explicit interface methods as `public` and fix up the names -- *maybe not exactly ideal* but it works

### DIFF
--- a/binder/Passes/InterfacesPass.cs
+++ b/binder/Passes/InterfacesPass.cs
@@ -64,7 +64,8 @@ namespace MonoEmbeddinator4000.Passes
 
             foreach (var method in methods)
             {
-                if (method.Name == "__getObject")
+                //NOTE: skip methods such as __getObject
+                if (method.IsImplicit)
                     continue;
 
                 var methodImpl = new Method(method)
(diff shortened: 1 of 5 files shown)


binder/Passes/InterfacesPass.cs
```diff
@@ -56,9 +56,17 @@ void CreateInterfaceImplementation(Class @interface)
...
-            var methods = @interface.Declarations.Where(d => d is Method).Cast<Method>();
+            var methods = new List<Method>(@interface.Declarations.OfType<Method>());
+            foreach (var baseInterface in @interface.Bases)
+            {
+                methods.AddRange(baseInterface.Class.Declarations.OfType<Method>());
+            }
+
             foreach (var method in methods)
             {
+                if (method.Name == "__getObject")
```
tritao: Think it might be better to check for `IsImplicit` here, which is only true in the case of synthesised/artificial methods, in case we have a need to add more such methods later to the interface.

PR #455, opened 2017-07-26
[android] support for AndroidNativeLibrary build action

### LINKED ISSUE #452: [Android] SQLite access failing from C# library
Accessing sqlite from the library is causing a linker failure, seen below:

> 31545-31980/com.s.t.android E/linker: library "/system/lib64/libsqlite.so" ("/system/lib64/libsqlite.so") needed or dlopened by "/data/app/com.s.t.android-2/lib/arm64/libmonosgen-2.0.so" is not accessible for the namespace: [name="classloader-namespace", ld_library_paths="", default_library_paths="/data/app/com.s.t.android-2/lib/arm64:/data/app/com.s.t.android-2/base.apk!/lib/arm64-v8a", permitted_paths="/data:/mnt/expand:/data/data/com.s.t.android"]
> 31545-31980/com.s.t.android I/mono-stdout: System.DllNotFoundException: /system/lib64/libsqlite.so
> 31545-31980/com.s.t.android I/mono-stdout:     at (wrapper managed-to-native) HIA2.C.DataAccess.SQLite.SQLite3Wrapper:Open (byte[],intptr&,int,intptr)

jonathanpeppers:
Problem:
- Xamarin.Android includes native libraries in an EmbeddedResource
named `__AndroidNativeLibraries__.zip`
- We need to extract the native libraries from this zip file into the
jni directory of the resulting AAR file
- We also need to strip `__AndroidNativeLibraries__.zip` from the
assembly along with `__AndroidLibraryProjects__.zip` to prevent wasted
file size

Changes:
- Needed to add Mono.Cecil from NuGet (unless IKVM.Reflection can edit
assemblies?)
- Included a C# integration test with a sample zip file created manually
- Included a sample C# SQLite class to test SQLitePCL.raw
- Included a Java integration test ensuring that SQLitePCL.raw works

Fixes #452


jonathanpeppers:
@tritao let me know if you think using Mono.Cecil from NuGet is fine.

tritao:
I think since this is using Cecil to re-save the assembly, the metadata tokens for the assembly types will change. That will be a problem once we start using metadata tokens as an optimization in Release mode for looking up the Mono types vs. solely using strings as we do now.

But we can solve that at a later time by doing this processing at the start and then re-reading the assembly tokens before generation.

tritao:
About Cecil, honestly I usually prefer submodules vs. relying on binary packages because it's easier to debug third party code if there's any problems. Anyway I think using the NuGet should be fine in this case.

PR #466, opened 2017-08-01
[tests] F# projects

jonathanpeppers:
- These are mainly smoke tests right now, as they seem to work fine on
Android/Java
- In the future, there might need to be specific F# test cases
- As usual, the SLN file changed a lot

#349


realvictorprm:
Awesome

jonathanpeppers:
@realvictorprm if you are more familiar with F# feel free to add F# code in this test project and see if it breaks anything. You can probably just add F# code and run `./build.sh -t Generate-Android-FSharp`. There is probably still a way to break IKVM.Reflection like on #349 

I have only used F# to the extend of using FAKE to build Xamarin projects, so I am probably an F# n00b.

baronfel:
Historically tail calls, lambdas, quotations, and inlining (especially around Statically Resolved Type Parameters) have been areas of pain. It would be worth making sure that any F# test cases include at least those :)

realvictorprm:
Yeah @jonathanpeppers,  I would be glad to test it!

PR #471, opened 2017-08-02
[android] fix for locating MSBuild on Windows

jonathanpeppers:
In updating the documentation for Embeddinator-4000, I found that `MSBuild.exe` was not found in `PATH` if trying to run Embeddinator as a post-build event inside Visual Studio.

A little research, and I found it is recommended to lookup the registry key `HKLM\SOFTWARE\Microsoft\MSBuild\ToolsVersions\4.0\MSBuildToolsPath` for finding `MSBuild.exe`.

This fix allows these commands to work inside VS:
```
set E4K_OUTPUT="$(SolutionDir)output"
if exist %E4K_OUTPUT% rmdir /S /Q %E4K_OUTPUT%
"$(SolutionDir)packages\Embeddinator-4000.0.2.0.80\tools\Embeddinator-4000.exe" "$(TargetPath)" --gen=Java --platform=Android --outdir=%E4K_OUTPUT% -c
```

### DIFF
--- a/binder/Utils/XamarinAndroid.cs
+++ b/binder/Utils/XamarinAndroid.cs
@@ -1,7 +1,8 @@
 using System;
-using System.Linq;
 using System.IO;
+using System.Linq;
 using CppSharp;
+using Microsoft.Win32;
 using Xamarin.Android.Tools;
 using static System.IO.Path;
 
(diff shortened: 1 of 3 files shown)


binder/Utils/XamarinAndroid.cs
```diff
@@ -144,5 +145,27 @@ public static string JavaSdkPath
...
         }
+
+        static Lazy<string> msBuildPath = new Lazy<string>(() =>
+        {
+            if (Platform.IsMacOS)
+                return "/Library/Frameworks/Mono.framework/Versions/Current/Commands/msbuild";
+
+            if (Platform.IsWindows)
+            {
+                using (var registryKey = Registry.LocalMachine.OpenSubKey(@"SOFTWARE\Microsoft\MSBuild\ToolsVersions\4.0"))
```
tritao: You can use this helper code from CppSharp for this: https://github.com/mono/CppSharp/blob/master/src/Core/Toolchains/MSVCToolchain.cs#L357

PR #474, opened 2017-08-07
[android] reference all assemblies for resource.designer.cs compilation

postromantic:
As there can be more than one Xamarin Android assemblies with resources as e4k input, so make sense to compile resource.designer.cs with all assemblies, not only main one.


postromantic:
@jonathanpeppers please have a look, i think it's the last issue i found in android resources area

PR #476, opened 2017-08-15
It should be able to support Linux, just add toolchain paths.

atsushieno:


atsushieno:
To Linux users: you need a working Cake to build it. Cake is broken on Linux at the moment and you need this fix https://github.com/cake-build/cake/pull/1749.

tritao:
Thanks for the fixes! If everything works, we should also update the docs stating Linux support.

PR #482, opened 2017-08-17
NameGenerator now honors the RegisterAttribute

zgramana:
Resolves mono/Embeddinator-4000#481 by allowing users to override
the generated objc class name by decorating their managed class
with the RegisterAttribute.

### DIFF
--- a/objcgen/NameGenerator.cs
+++ b/objcgen/NameGenerator.cs
@@ -53,6 +53,12 @@ namespace ObjC {
 
 		public static string GetObjCName (Type t)
 		{
+			if (t.IsClass) {
+				var ra = t.GetCustomAttributesData ().SingleOrDefault (a => a.AttributeType.Name.Equals ("RegisterAttribute"));
+				if (ra != null) {
+					return ra.ConstructorArguments.First ().Value.ToString ();
+				}                
+			}
 			return t.FullName.Replace ('.', '_').Replace ("+", "_");
 		}
 


tritao:
whitelist

dalexsoto:
Hello! Glad to see you contributing!! I am not entirely sold on reusing 'RegisterAttribute' for this because it could potentially interfere with the Xamarin registrar/runtime. Maybe we'll need embeddinator specific attributes.

zgramana:
Thanks, old friend! FWIW, I chose to re-use it since the official Xamarin docs state that RegisterAttribute is now used solely to customize the objc class name emitted by the XI generator. Adding a second attribute with the same responsibility would be a DRY violation. If I have a XI assembly using them I'd want it to Just Work. Having two attributes decorating the same class with the same values would be mildly awkward.

tritao:
FWIW we also rely on `[Register]` for this purpose in the Java backend, a behaviour we inherit from Xamarin.Android Java interop implementation.

dalexsoto:
Awesome, agree on the intention of reusing it, if @spouliot and @rolfbjarne are ok with this I am too!

rolfbjarne:
I'm fine with using `[RegisterAttribute]`. However I think you should be more defensive (it might not have an argument (`[Register]`), in which case your code will throw an exception. I'd do something like this, except that I'd issue a warning instead of an error if some assumption (maybe someday there will be 3 arguments?) breaks:

https://github.com/xamarin/xamarin-macios/blob/a001cace6f1e77646cd63e0119ab344f3935851b/tools/common/StaticRegistrar.cs#L1109-L1143

@lemonmojo note that this code only does a string comparison for the attribute name; this means that you don't have to reference Xamarin.iOS/Xamarin.Mac, you can define your own `RegisterAttribute` class.

zgramana:
@rolfbjarne I like it. I'll update the PR. With respect to the other params, like `IsWrapper`, should the switch ignore them, log them at verbose log level, or some third option I'm not able to think of?

rolfbjarne:
@zgramana Just ignore the `IsWrapper`field (no logging needed), it doesn't matter for the embeddinator (and I think you can ignore the `SkipRegistration` field as well, at least for now, but this one you might want to log, since users might want to try to use it for the obvious purpose). You might want to log other (unknown) fields as well.

zgramana:
I just pushed a refactor based on @rolfbjarne feedback for review.

zgramana:
Actually, hold off for another commit forthcoming. Apologies for the false start.

zgramana:
Okay, _now_ it's ready for review.

lemonmojo:
@zgramana @rolfbjarne Thx for clarifying that there's no external dependency required for this. In this case the PR sounds good to me.

I'd still like to be able to also rename specific properties/methods/etc. and exclude certain classes/properties/etc. from being generated but I guess that's fuel for another Issue/PR.

lemonmojo:
@zgramana Just noticed a small typo in the updated documentation: ```Users already familiar with the RegisterAttribute might expect SkipRegistration to function ---> is <--- it does in Xamarin.iOS or Xamarin.Android. However, this is unsupported and the value of SkipRegistration will be ignored.```

docs/errors.md
```diff
@@ -217,6 +217,53 @@ Selectors on the [NSObjectProtocol](https://developer.apple.com/reference/object
...
 
+<h3><a name="EM1060"/>Invalid `RegisterAttribute` found on class `T`. Expected the first constructor parameter to be of type `string` but found type `U` instead.</h3>
+
+This is a **warning** that the class `T` was decorated with a `RegisterAttribute`, indicating the author wishes to rename the name of generated objective-c class.
+
+When either one or two arguments are passed into the constructor, the first must be of type `System.String` but in this case, some other type was provided instead. For this to happen, the `RegisterAttribute` did not come from Xamarin.iOS or Xamarin.Mac.
+
+If an argument was not also passed for named parameter "Name" then the class will be named according to the default convention.
+
+<h3><a name="EM1061"/>Invalid `RegisterAttribute` found on class `T`. Expected a constructor with either 1 or 2 parameters but found {argsCount} instead..</h3>
```
rolfbjarne: Typo: two periods at the end of the sentence.

docs/errors.md
```diff
@@ -217,6 +217,53 @@ Selectors on the [NSObjectProtocol](https://developer.apple.com/reference/object
 
 Note: The list of reserved selectors will evolve with new versions of the tool.
 
+<h3><a name="EM1060"/>Invalid `RegisterAttribute` found on class `T`. Expected the first constructor parameter to be of type `string` but found type `U` instead.</h3>
+
+This is a **warning** that the class `T` was decorated with a `RegisterAttribute`, indicating the author wishes to rename the name of generated objective-c class.
```
rolfbjarne: Minor: use proper casing `Objective-C` instead of all lowercase (there are also more cases below).

docs/errors.md
```diff
@@ -217,6 +217,53 @@ Selectors on the [NSObjectProtocol](https://developer.apple.com/reference/object
...
+
+However, two conflicting values were given for 'Name': one as a positional argument, the other as a named parameter. For example, consider: `[Register("myFoo", Name = "foo")]`.
+
+The value of the named parameter, 'foo' will be used instead of value provided to the positional parameter: 'myFoo'. If that is not the value intended, remove the named argument.
+
+<h3><a name="EM1063"/>Invalid `RegisterAttribute` found on class `T`. Named parameter `SkipRegistration` is not supported and will be ignored.</h3>
+
+This is a **warning** that the class `T` was decorated with a `RegisterAttribute`, indicating the author wishes to rename the name of generated objective-c class.
+
+Users already familiar with the `RegisterAttribute` might expect `SkipRegistration` to function is it does in Xamarin.iOS or Xamarin.Android. However, this is unsupported and the value of `SkipRegistration` will be ignored.
```
rolfbjarne: I'm not sure `SkipRegistration` is applicable to Xamarin.Android (I don't know anything about XA). Did you mean Xamarin.Mac instead of Xamarin.Android?

zgramana: Yes, I did, thanks for catching that. My fingers seem to have their own L4 cache…

zgramana:
@rolfbjarne I've pushed up the copy edits you and @lemonmojo found.

Unfortunately I'm going to have to leave the tests to someone else as I have to attend to business commitments. Hope this is PR is still helpful.

PR #489, opened 2017-08-23
Extend unmangle type name method to respect F# naming features.

realvictorprm:
Would be worth to add an test for this...

~In short, we assume that _no white-spaces_ are allowed in _declaration names_. Therefore every white-space is replaced with an underscore.~

F# allows to make really strange variable names through using two backticks:
```fsharp
let ``supercrazyvarname 2345$"§"&/%43)=%4ß?\`` = 1
```

Signed-off-by: realvictorprm <mueller.vpr@gmail.com>

### DIFF
--- a/binder/Generators/AstGenerator.cs
+++ b/binder/Generators/AstGenerator.cs
@@ -6,6 +6,7 @@ using CppSharp.AST;
 using CppSharp.AST.Extensions;
 using CppSharp.Generators;
 using IKVM.Reflection;
+using System.Text.RegularExpressions;
 
 namespace Embeddinator.Generators
 {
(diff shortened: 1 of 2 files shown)


realvictorprm:
@tritao this is ready

binder/Generators/AstGenerator.cs
```diff
@@ -516,7 +517,7 @@ QualifiedType VisitType(IKVM.Reflection.Type managedType)
 
             return new QualifiedType(type, qualifiers);
         }
-
+
```
realvictorprm: ups, this shouldn't be here

realvictorprm: fixed

PR #491, opened 2017-08-25
[WIP] Replacing CAKE + NuGet with FAKE + Paket 

realvictorprm:
CAKE and NuGet aren't the best solution as build system.

FAKE and Paket ease the dependency management and provide a cleaner and nicer syntax for writing build files thanks to F#. Also with FAKE there isn't anything additional required than Visual Studio with F# support for editing the build files because F# supports script files _per se_.

This is work in progress, I hope to finish this soon.

Signed-off-by: realvictorprm <mueller.vpr@gmail.com>


realvictorprm:
I need to check how the NUnit tool works here.
Also travis and appvoyer need to be changed slightly I guess.

build.fsx
```diff
@@ -0,0 +1,159 @@
...
+
+    let private premake = 
+        let exe = 
+            if isWindows then "premake5.exe" 
+            elif isMacOS then "premake5-osx"
+            elif isLinux then "premake5"
+            else 
+                traceError "Unsupported operating system detected!"
+                ""
+        Path.Combine("external", "CppSharp", "build", exe)
```
realvictorprm: replace with `@@`

jonathanpeppers:
Hi @realvictorprm,

Just asking, but could your efforts be better used someplace else? (in Embeddinator)

Redoing the build system doesn't really gain that much, even _if_ F#/FAKE _might_ be a factor of two better than C#/Cake. It is still _just a build_, and not giving end-users any new features.

Is there something else F#-related that could be worked on instead? I can't imagine redoing the build is that exciting anyway.

realvictorprm:
Hi @jonathanpeppers,

thanks for your feedback. I agree that the impact on end-user experience is only less here. However for us developer the impact is high because it should ease our life and the maintainability of the build system (For me it reduces the pain to write build scripts and safes time).

I've started this after I noticed that the build system still does not work fine enough for me during adding Tests.
 
----

I'm already aiming for adding more F# tests currently for #489. Next week I plan adding support for generics to the C generator because I need it for production. This increases support for F# too and might include some F# specific stuff aswell.

I hope to present this soon enough in a good state so you see which improvements we have.

tritao:
Hey @realvictorprm, as I told you we just stabilized the switch to Cake, we're not switching build system again, even if it provides some advantages.

Let's focus on something more productive instead!

PR #525, opened 2017-10-24
Update Xamarin.Android.Tools to latest

jonathanpeppers:
[Xamarin.Android.Tools](https://github.com/xamarin/xamarin-android-tools) is going to be shipped in a future version of Xamarin.Android. They've made several API changes up to this point, such as making `AndroidSdk` non-static, improved logging events, and removed `MonoDroidSdk`.

We should keep Embeddinator on the latest Xamarin.Android.Tools, as it has several fixes, and will continue to get fixes as more projects rely on it.

I included the source for the now absent `MonoDroidSdk` in `external/MonoDroidSdk`, as it was used for locating Xamarin.Android.

Other changes:
- `TargetSdkVersion` should be 24 to match "v7.0" - down the road we could add a command line switch for users to specify API level or bump to 26 and "v8.0"
- Added some `.gitignore` for Windows users
- Fixed paths in `premake.lua` for Xamarin.Android.Tools and Xamarin.MacDev

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,8 +65,10 @@ xcuserdata/
 *.dSYM.zip
 *.dSYM
 *.nupkg
+*.bak
 .cake/**
 !.cake/packages.config
 .vscode/
+.vs/
 Resource.designer.cs
 external/Xamarin.Android
--- a/binder/Compilation.cs
+++ b/binder/Compilation.cs
@@ -627,7 +627,7 @@ namespace Embeddinator
             var monoPath = Path.Combine(MonoSdkPath, "include", "mono-2.0");
             var name = Path.GetFileNameWithoutExtension(Project.Assemblies[0]);
             var libName = $"lib{name}.so";
-            var ndkPath = AndroidSdk.AndroidNdkPath;
+            var ndkPath = XamarinAndroid.AndroidSdk.AndroidNdkPath;
 
             foreach (var abi in new[] { "armeabi", "armeabi-v7a", "arm64-v8a", "x86", "x86_64" })
             {
--- a/binder/Utils/NdkUtils.cs
+++ b/binder/Utils/NdkUtils.cs
@@ -1,9 +1,10 @@
-﻿using CppSharp;
-using System;
+﻿using System;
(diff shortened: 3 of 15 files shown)


jonathanpeppers:
Hmm, the failure looks like it is related to not finding an Android SDK. I guess Travis doesn't have it?

Due to the code here: https://github.com/xamarin/xamarin-android-tools/blob/master/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs#L24

I guess this is an issue as `Generate-Java` should be able to work without an Android SDK.

tritao:
Looks like we are using `XamarinAndroid.JavaSdkPath` to get the Java SDK path even in non-Android code path. The issue is that this now seems to trigger Android SDK lookup inside Xamarin.Android.Tools:

```
Unhandled Exception:
System.TypeInitializationException: The type initializer for 'Embeddinator.XamarinAndroid' threw an exception. ---> System.InvalidOperationException: Could not determine Android SDK location. Please provide `androidSdkPath`.
 at Xamarin.Android.Tools.AndroidSdkInfo..ctor (System.Action`2[T1,T2] logger, System.String androidSdkPath, System.String androidNdkPath, System.String javaSdkPath) [0x00082] in /home/travis/build/mono/Embeddinator-4000/external/Xamarin.Android.Tools/src/Xamarin.Android.Tools.AndroidSdk/AndroidSdkInfo.cs:28
 at Embeddinator.XamarinAndroid..cctor () [0x000ae] in /home/travis/build/mono/Embeddinator-4000/binder/Utils/XamarinAndroid.cs:26
```

jonathanpeppers:
Yeah the code that is throwing an exception here doesn’t seem like it should behave that way. I’m going to ask today and see if I should fix Xamarin.Android.Tools or maybe add a new ctor that doesn’t throw.

jonathanpeppers:
@tritao I think this one is good. It seems everyone wants `AndroidSdkInfo` to throw, but if you are using only java `Lazy<T>` fixes this like you suggested.

PR #540, opened 2017-11-08
[objc] Extension use support

equinox2k:
This update contains the needed code to allow user to enable "Require Only App-Extension-Safe API" which is needed in app extensions...

see ...

[https://developer.apple.com/library/content/documentation/General/Conceptual/ExtensibilityPG/ExtensionScenarios.html#//apple_ref/doc/uid/TP40014214-CH21-SW1](https://developer.apple.com/library/content/documentation/General/Conceptual/ExtensibilityPG/ExtensionScenarios.html#//apple_ref/doc/uid/TP40014214-CH21-SW1)

> Make sure your embedded framework does not contain APIs unavailable to app extensions, as described in Some APIs Are Unavailable to App Extensions. If you have a custom framework that does contain such APIs, you can safely link to it from your containing app but cannot share that code with the app’s contained extensions. The App Store rejects any app extension that links to such frameworks or that otherwise uses unavailable APIs.

> To configure an app extension target to use an embedded framework, set the target’s “Require Only App-Extension-Safe API” build setting to Yes. If you don’t, Xcode reminds you to do so by displaying the warning “linking against dylib not safe for use in application extensions”.


objcgen/driver.cs
```diff
@@ -746,7 +754,8 @@ public int Compile ()
 							mmp.Append ("--debug ");
 						mmp.Append ("-p "); // generate a plist
 						mmp.Append ($"--target-framework {GetTargetFramework ()} ");
-						mmp.Append ($"\"--link_flags=-force_load {Path.GetFullPath (sdk_output_file)}\" ");
+                        string extensionFlag = Extension ? "-fapplication-extension" : "";
```
tritao: Can you fix indentation here?

equinox2k:
@tritao indentation fixed now

PR #547, opened 2017-11-24
Ensure that not internet connection is needed.

mandel-macaque:
The default reader settings tries to download the DTD from the network, that in our case, makes no sense.


tritao:
@spouliot `objc` branch is fully merged into `master` branch. I suggest you start working on `master` going forward.

mandel-macaque:
Closing since it was easier to create a new PR. @spouliot you can find the new one here: https://github.com/mono/Embeddinator-4000/pull/551

PR #553, opened 2017-11-29
Logo - Fixed kerning between A and T

jongalloway:
Re-centered logo after fixing kerning as it's a little less wide now


tritao:
Thanks

PR #555, opened 2017-12-12
Add required Mono.TextTransform dependency to packages.config

salockhart:
When building, build will fail if this package is not present.  Adding a dependency to the package in nuget fixes the issue.

This issue was initially found on macOS High Sierra, Visual Studio for Mac Community Edition


tritao:
Thanks for the PR @salockhart. 

Can you add it here too?

https://github.com/mono/Embeddinator-4000/blob/master/binder/premake5.lua#L11

PR #556, opened 2017-12-13
Add required Mono.TextTransform dependency to premake5.lua

salockhart:


jonathanpeppers:
@salockhart it looks like this change failed to build

I think we are talking about dropping premake anyway, is this still needed on latest master?

PR #569, opened 2018-01-15
Do not lookup Java SDK path on global Cake scope.

tritao:
Fixes https://github.com/mono/Embeddinator-4000/issues/564.


Cronan:
Thanks, I'll test this in my environment

Issue #571, opened 2018-01-18
Java: unsafe pointer types are bound as pointee types


atsushieno:
If you run e4k against mscorlib.dll, there is UIntPtr.ToPointer() which returns void*, and it results in:

        public void toPointer() {
            void __ret = mscorlib.Native_mscorlib.INSTANCE.System_UIntPtr_ToPointer(__object);
            mono.embeddinator.Runtime.checkExceptions();
            return __ret;
        }

PR #572, opened 2018-01-18
JavaTypePrinter: return JNA Pointer for any pointer type

atsushieno:
I'm almost sure that there should be more appropriate way to return a TypePrinterResult (especially with TypeMap) but this at least generates non-void return code.

This should fix https://github.com/mono/Embeddinator-4000/issues/571 (I still cannot compile the entire mscorlib.dll due to other issues).


atsushieno:
I guess the fix was completely wrong. There should be other way to achieve that.

tritao:
Hey @atsushieno sorry for lack of feedback before.

Fix is fine, but you don't need any of type map, you can just return the JNA type as string.

I will add this with a test case.

tritao:
Actually it's a bit more complicated due to ref/out types which are implemented as pointers in the AST too.

PR #576, opened 2018-01-19
[Objc] Filter devices that are not availabe for debug. Fixes #575

### LINKED ISSUE #575: Objc branch tests fail when Wifi devices are available
In the cases we a developer has a device present in the same WiFi not enabled for debugging the following happens:

xcodebuild: error: Unable to find a destination matching the provided destination specifier:
        { platform:iOS, id:bd3fd054136767445ff47819f65b74867a48e591 }

when the iOS device is not available. Filtering them using IsUsableForDebugging node would be a fix to ensure that we do not get the error.

mandel-macaque:


mandel-macaque:
Landing PR since the failures are unrelated to this changes.

Issue #573, opened 2018-01-18
Java: long constants result in compiiation error


atsushieno:
If you try to bind mscorlib.dll to generate Java binding, it results in compilation error due to too large number for integers:

        /home/atsushi/Desktop/TextToSpeechPlugin-master/e4k/android/mscorlib/system/diagnostics/tracing/EventKeywords.java:17: error: integer number too large: 562949953421312
    public static final EventKeywords MicrosoftTelemetry = new EventKeywords((long)562949953421312);

See https://stackoverflow.com/questions/3757763/integer-number-too-large-error-message-for-600851475143 for solution. Though I couldn't quickly find which code to fix.

PR #579, opened 2018-01-25
Java long constant requires 'l' suffix

atsushieno:
This fixes https://github.com/mono/Embeddinator-4000/issues/573


tritao:
Thanks for PR, but fixed the bug in a more generic way.

Issue #580, opened 2018-01-25
Java: properties easily generate conflicts with methods, even in mscorlib types


atsushieno:
e4k generates properties with "get" prefix, unlike C# internal do ("get_"). This API design easily results in name conflicts in the class libraries.

For example, in mscorlib.dll there is [System.Runtime.Remoting.Metadata.W3cXsd2001.SoapId](https://msdn.microsoft.com/en-us/library/system.runtime.remoting.metadata.w3cxsd2001.soapid_members%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396
) has XsdType static property and GetXsdType() method. They are both generated as getXsdType(). This happens everywhere in this namespace.

tritao:
We used to generate with `get_` too but this was improved recently to better match Java conventions but looks like some edge cases were missed.

PR #585, opened 2018-01-26
[java] Fixed generation conflict with properties and methods

tritao:
Fixed generation conflict with properties and methods starting with get/set.

This was a regression caused by https://github.com/mono/Embeddinator-4000/pull/542

Fixes https://github.com/mono/Embeddinator-4000/issues/580.

### DIFF
--- a/binder/Passes/CheckDeclarationsPass.cs
+++ b/binder/Passes/CheckDeclarationsPass.cs
@@ -16,6 +16,11 @@ namespace Embeddinator.Passes
             // Remove get_ or set_ from names in case they are properties.
             var name = decl.Name.TrimStart("get_").TrimStart("set_");
 
+            // Do the same for method names starting with get or set since they
+            // might conflict with properties.
+            name = decl.Name.TrimStart("get").TrimStart("set").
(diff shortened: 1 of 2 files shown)


binder/Passes/CheckDeclarationsPass.cs
```diff
@@ -16,6 +16,11 @@ static string GetBaseName(Declaration decl)
             // Remove get_ or set_ from names in case they are properties.
             var name = decl.Name.TrimStart("get_").TrimStart("set_");
 
+            // Do the same for method names starting with get or set since they
+            // might conflict with properties.
+            name = decl.Name.TrimStart("get").TrimStart("set").
+                             TrimStart("Get").TrimStart("Set");
+
```
atsushieno: Does this work with "GetGetFoo()" ? For example, PropertyInfo.GetGetMethod() and GetSetMethod() https://msdn.microsoft.com/en-us/library/system.reflection.propertyinfo.getgetmethod%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396

tritao: I will check.

PR #603, opened 2018-02-21
Corrected various Visual Studio 2017 build issues

optikos:
When building with the current Visual Studio 2017 Preview, I observed the following issues and made the following corrections:

1) objcgen's objcgen.csproj incorrectly invoked ./generate_version.sh via forward slash, making . be parsed as a command, which in turn was unrecognized.  Removed spurious ./

2) objcgentest was missing required references to Emdeddinator-4000 and to objcgen.  Added both.

3) (Some people might have legacy installations of .NET 4.5 SDK; I do not.) .NET 4.5 SDK is no longer downloadable from Microsoft (or any other legit website) without an Enterprise MSDN account.  On the current Windows 10 Insider Preview slow-ring, Windows 8 SDK ended prematurely due to a nondescript error that referred to the log files, so installing Windows 8 SDK failed to install .NET 4.5 SDK.  Global Assembly Cache has commingled versions of .NET, causing .NET 4.5 SDK or Targeting Pack to not be found overtly (although it was installed for Visual Studio 2017 Preview).  Following the guidance given in the error message, eliminated the GAC-based comingled mixture of releases of .NET by changing all projects to use .NET 4.6.1, following CppSharp's lead.  Declared so in all .csproj files that formerly referred to .NET 4.5 or .NET 4.6.

4) tests/managed/managed-macos-full and tests/managed/managed-macos-system were not like managed-macos-modern in that the former 2 did not overtly declare that Xamarin.Mac was the edition of (Mono-).NET to use.  Declared so in those 2 .csproj files.

5) tests/fsharp-shared project does not load.  I did not fix that.  That is the sole remaining build error and sole remaining target not built.

I did not execute any regression tests, because frankly I do not understand how to invoke them properly.  Please advise.

The build/projects/CppSharp.AST.csproj file's CRLF line termination inadvertently got changed to LF.  I changed it back, but now it looks like every line changed here at GitHub.


jonathanpeppers:
Hi @optikos this PR has some good fixes in it, but I'm afraid it's doing a lot at once. (It's also been a while, sorry!)

Here are a few comments:
- They are using `premake` in this repo to generate lots of these csproj files. So it becomes a little moot to modify these, since the build re-generates them. Run the Cake build targets (`./build.ps1`) to see how this works
- The macOS-specific commands probably need to have an MSBuild condition to skip them on Windows. It doesn't make sense to remove the `./` if the command likely won't work anyway.
- `TargetFrameworkVersion` is being changed to 4.6.1 in various spots, not sure if we need that.

Maybe sending a new PR with the latest from master (this PR conflicts now), and make less changes at once? You might get along better by hand editing the project files instead of using an IDE.

optikos:
> TargetFrameworkVersion is being changed to 4.6.1 in various spots, not sure if we need that.

Then you do not understand how the Global Assembly Cache works.  Once you introduce a later version of .NET SDK (or Targeting Pack) on your build machine (e.g., 4.6, 4.6.1), then you are using that later version, not the 4.5.  The later version eclipses the earlier version for same-named identifiers.  If you have .NET SDK (or Targeting Pack) 4.5, 4.6, 4.6.1, 4.6.2, and 4.7 installed on your build machine, then you are building a different version than someone else who has a different mixture of .NET SDK (or Targeting Pack) versions installed.

You absolutely need to make all portions of Embeddinator-4000 build off from the same single .NET SDK.  When building Embeddinator-4000, the compiler error that not having .NET SDK (or Targeting Pack) 4.5 installed on your build machine even says so.

(Neither .NET SDK 4.5 nor .NET Targeting Pack 4.5 are available publicly anymore for download, nor are they available at the cheaper end of the MSDN subscriptions.  I cannot speak for the highest-end MSDN subscriptions.  You likely have .NET SDK 4.5 installed on your build machine, but I do not and absolutely cannot, so I get different severe build errors than you get.  These build errors effectively say that everything needs to be at least 4.6.1 across the board.)

jonathanpeppers:
If you want to target 4.6.1 and get all the projects on the same version, let's do that in a separate smaller PR. You will need to update the premake files as well, and make our cake script works to generate the project files.

> Then you do not understand how the Global Assembly Cache works. 

Let's just try to keep things light.

optikos:
I will try to allocate time to recreate this work piecemeal as separate PRs.  I will try to figure out cake to generate the correct content in the project files; that is new to me.

Generally, when submitting a pull-request (or the analogue in other version-control systems) for correcting multiple build issues (each of which causes the build to fail), I have always been under the opinion that the minimum viable measuring stick of success for accepting a valid PR is:  does it build correctly now?  (Or better yet, does it build correctly now and pass most/all of its regression test-suite?)  Because there were multiple problems in building Embeddinator-4000 on Visual Studio 2017 without .NET SDK 4.5 anymore, (minus the accidental CRLF versus LF transform in GNU Emacs in that one file) this was the minimal set of changes needed to get Embeddinator-4000 to successfully build in Visual Studio 2017 without no-longer-available .NET SDK 4.5.  When I resubmit this PR as multiple PRs, no one PR in isolation will fix Embeddinator-4000's build problems; all of the separate PRs will be needed together acting in concert.  So in effect, it will be hardly any different than a single PR (except for the accidental CRLF-LF-CRLF spurious all-lines-changed in that one file and except for the changes to the upstream cake files).  Isn't it inappropriate to accept a PR that transforms a code base from not-building to still-not-building?  Isn't the minimum for PR acceptance to transform a code base from not-building to oh-now!-it-builds?

Perhaps there is a better way of building all the Microsoft Windows-targeted portions of Embeddinator-4000 than loading the solution file and telling the IDE to compile all projects in that solution.  (Apparently there is, given the cake prerequisite.)  I just thought that open-solution-and-let-'er-rip was the intended build of Embeddinator-4000.

Reproducibility:
I suspect that the disconnect (in addition to the cake generation of project files different than what is checked in* as XML files) is that nearly no one is building Embeddinator-4000 on a freshly-configured pristine Windows 10 build machine with a freshly-installed pristine Visual Studio 2017.  Everyone else evidently already has .NET SDK 4.5 (or the Targeting Pack 4.5 analogue thereof) already installed from years ago, bypassing this entire GAC issue of silently-implicitly mixing & matching different .NET SDKs versions as the compiler error rather verbosely taught in you've-got-a-quite-nasty-multiple-assembly-versions-in-GAC-problem-here-and-here-is-the-customary-solution style of speaking.  It was the most-revealing mini-tutorial embedded into a compiler error that I have perhaps ever received in 4 decades of writing software.

* I am a little confused why unmodified cake files in Embeddinator-4000 version X wouldn't merely regenerate the exact same content as what is checked in as XML project files in Embeddinator-4000 at the same version X.  I.e., why are the generated XML project files out of date with respect to their upstream cake source code in GitHub at the same instant in time?  Clearly, merely regenerating the project files shouldn't magically cause all my Visual-Studio-2017 and .NET-SDK-4.5-no-longer-available build issues to go away (without enacting the same changes that I handcrafted in the various project & configuration files but that would arrive at nearly the same too-big PR by merely traveling along a different cake-based road instead of an XML-based road) because the checked-in XML project files would have already been the exact same as the freshly-generated project files at any release X of Embeddinator-4000 in GitHub.

jonathanpeppers:
A lot of these decisions were made before I started helping on this project. Maybe @tritao can fill in the gaps?

Keep in mind that most development is done on Mac--this probably explains any of the oddities you are seeing. In my opinion, I don't think that premake helps us much here, I would rather just commit the project files myself.

Our CI system also builds every commit on Windows (AppVeyor), and that has been succeeding for some time now. Perhaps alot of these issues are just IDE-related in VS 2017? Are you getting the same problems building command line when running the cake script?

Zomb:
Here is my small cheat-sheet that I use when setting up e4k for use on Windows. It's on a pristine VM, but it works for me!

install windows
install Mono 5.0 32 bit
install Java 1.8 JDK
install Git

clone repo
get submodules

open powershell as administrator and run
Set-ExecutionPolicy RemoteSigned
to enable scripts

Install Visual Studio with the following Components: .NET Desktop Development, Mobile Development
Open VS -> Update Android SDK -> If you don't do this you can get an SDK not found error.

Change android-sdk-windows to android-sdk in the AndroidSdkWindows.cs file (I believe @jonathanpeppers has fixed this, so it's probably not needed)

Run the following script to compile e4k
.\build.ps1 -t Generate-Android -v diagnostic

optikos:
Oh!  Mono instead of Microsoft .NET!  That's a difference that would make a difference.

tritao:
Hey, sorry for not giving some feedback earlier on this PR.

The reason I originally chose Premake and generated project files is precisely because of these kinds of issues with regards to mismatched SDKs that can happen on Windows and MSBuild projs.

I think it does not make sense to continue using it because this project is not relying on unmanaged code, and most regular .NET devs are not aware of how it works or how to use it.

So I think it's fine to just get rid of it and start using regular committed MSBuild projs. The only thing to be aware of is that CppSharp project files are generated using Premake, because of the need to build unmanaged code in a cross-platform way.

Regarding Windows CI, I am not sure what you propose. Our AppVeyor config is pretty "stock", just a regular VS2017 image in which we install Mono.

@optikos Try `./build.ps1 Generate-Project-Files` which will hopefully give you working project files.

And regarding your previous comment, we do not use Mono to build Embeddinator on Windows. We just rely on the Mono assemblies as part of the actual embedding process.

optikos:
> Regarding Windows CI, I am not sure what you propose. Our AppVeyor config is pretty "stock", just a regular VS2017 image in which we install Mono.

I propose that all maintainers of Embeddinator-4000 take a long hard look at .NET SDK 4.5's lack of public availability.  (Apparently) using .NET SDK 4.5 in your continuous integration builds is not representative of reality for people (outside of Microsoft) downstream who now lack the ability ever acquire .NET SDK 4.5 ever again.

jonathanpeppers:
@optikos I'm a bit confused why you are having trouble getting .Net 4.5 in VS 2017.

You just need to check these boxes from the VS installer:
image

I can create .Net 4.5 console apps and run them, just tested it.

Am I misunderstanding your issue? I have Visual Studio 2017 15.6.3 at the moment.

optikos:
@jonathanpeppers 
Yes, I have those installed from Visual Studio 2017 15.6.4, as you describe with the exception that I also have all the 4.6.2, 4.7, and 4.7.1 checkboxes also checked.  Upon building via the instructions‡ on the README.md "Building … Open the solution file Embeddinator-4000.sln with Visual Studio … and press F7.", I receive the following error in unmodified Embeddinator-4000 straight from GitHub cloned via:  `git clone --recurse-submodules https://github.com/mono/Embeddinator-4000.git`

`15>C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets(1126,5): error MSB3644: The reference assemblies for framework ".NETFramework,Version=v4.5" were not found. To resolve this, install the SDK or Targeting Pack for this framework version or retarget your application to a version of the framework for which you have the SDK or Targeting Pack installed. Note that assemblies will be resolved from the Global Assembly Cache (GAC) and will be used in place of reference assemblies. Therefore your assembly may not be correctly targeted for the framework you intend.`

In the pull-request earlier work, I was able to eliminate this error only by eliminating all dependency on .NET 4.5 Targeting Pack in all assemblies that had targeted the defunct‡‡ .NET 4.5.

‡ I have not yet invested the time needed to learn the command-line method of building that is more complex than `make <enter-key>` or `autoconf; make <enter-key>` or using the .sln files within Visual Studio.  Overcoming the deviations from the Principle of Least Surprise usually requires at least tens of minutes, or perhaps hours or days, which then must be scheduled in.  There must exist some significant complexity that must be learned/teased-apart if building is more complex than `make <enter-key>` or `autoconf; make <enter-key>` or using the .sln files within Visual Studio.  I have not yet scheduled the time needed for such a cycle of learning.  It seems that the README.md should be updated to say how to build from the command-line as per Zomb's technique above instead of indicating that load-.sln-press-F7 is the normative build technique.

Or in fewer words, I am not pulling this GAC stuff out of thin air (or any other variant of being a pain in the patoot).  Co-mingled versions of .NET in the GAC appear to be the root-cause of this error that actually occurs in unmodified Embeddinator-4000 straight from GitHub using the README.md's build instructions.  In preparing the earlier pull-request's work, I merely compliantly obeyed what Microsoft's (quite-informative-tutorial) error message clearly (and wisely?) instructed to do:  retarget the 4.5-targeted assemblies to all be the same more-modern version of .NET.

‡‡ As announced over 2 years ago (!), support has ended on .NET 4.5 as of 12 January 2016:
https://blogs.msdn.microsoft.com/dotnet/2015/12/09/support-ending-for-the-net-framework-4-4-5-and-4-5-1

jonathanpeppers:
Have you tried running the [Cake](https://cakebuild.net/) build? I mentioned it, but this thread is quite long.

In powershell:
```
.\build.ps1
```
Cake usually includes two bootstrappers (one for Windows and one for Mac).

Docs mention it [here](https://github.com/mono/Embeddinator-4000/blob/master/docs/Contributing.md#cake).

A PR updating all the project files to 4.6.x would be great! Or perhaps removing premake?

PR #667, opened 2018-05-01
[windows] .gitattributes for proper line endings settings

jonathanpeppers:
Context: https://help.github.com/articles/dealing-with-line-endings/

If you work on this repo on Windows, you quickly see the `.gitattributes` are not setup right for development on Windows.

1. Certain files had mixed line endings, which causes Visual Studio to prompt to adjust.

2. Running `premake` yields Windows line endings on Windows, and so every `*.csproj` file appears to have changes.

`xamarin-android` has a well-used `.gitattributes` file, so I used it as a starting point and added an entry for `*.cake` files.

The way it works (as I understand it):
- `git` stores everything with unix line endings
- on checkout, clients convert the line endings according to settings in the `.gitattributes` file

So, for example:
- All `*.csproj` files will have Windows line endings, as Visual Studio expects (both Mac and Windows)
- `*.cs` files will have Windows line endings on Windows, and unix line endings on Mac


jonathanpeppers:
Here is an example where we went through this process on `java.interop`: https://github.com/xamarin/java.interop/pull/277

Comment [here](https://github.com/xamarin/java.interop/pull/277#issuecomment-376951623) where we compared to what was already setup in `xamarin-android`.

.gitattributes
```diff
@@ -0,0 +1,32 @@
+# This file mainly controls line ending conversion behaviour, if
+# the user has the setting core.autocrlf true.
+
+# The meaning of the attributes is a little odd
+# -crlf means DO NOT convert line endings
+# crlf means CONVERT to lf in the repo & Linux/Mac, crlf on Windows
```
rolfbjarne: This comment is not correct according to the git documentation (https://git-scm.com/docs/gitattributes), it should be something like:

```
# crlf means CONVERT to lf in the repo, but always checkout using crlf on all platforms
```

PR #674, opened 2018-05-24
Avoid copying const string constants to the stack.

jaykrell:


objcgen/methodhelper.cs
```diff
@@ -74,7 +74,7 @@ public void WriteMethodLookup ()
 			implementation.Indent++;
 			implementation.WriteLine ($"__method = mono_get_method (__{AssemblySafeName}_image, 0x{MetadataToken:X8}, {ObjCTypeName}_class);");
 			implementation.WriteLineUnindented ("#else");
-			implementation.WriteLine ($"const char __method_name [] = \"{ManagedTypeName}:{MonoSignature}\";");
+			implementation.WriteLine ($"static const char __method_name [] = \"{ManagedTypeName}:{MonoSignature}\";");
```
rolfbjarne: Will this generate the same native code as if the string constant had been passed directly to the method call?

i.e.:

```c
implementation.WriteLine ($"__method = mono_embeddinator_lookup_method (\"{ManagedTypeName}:{MonoSignature}\", {ObjCTypeName}_class);");
```

jaykrell: Another way is to pass the literal directly.
There are tradeoffs either way.
With msvc:
Literals can be writable or not, depending on version/flags.
Literals can be single-instanced or not, depending on version/flags.
static const will be not writable and not single-instanced.
Ideal is readonly and single instanced.
Non-static const at least with msvc will copy to stack.
Though I have seen clang optimize similar, kinda surprising.
Perhaps literals w/o temporary are best.
Sometimes people will also assign the literal to a pointer.
That is a wasted assignment but only of a pointer, so minor.

PR #725, opened 2019-04-10
[objcgen] Fix shellcheck warning.

rolfbjarne:
Fixes this warning:

    In thin-framework.sh line 70:
    		let C++ || true
                    ^------^ SC2219: Instead of 'let expr', prefer (( expr )) .

    For more information:
      https://www.shellcheck.net/wiki/SC2219 -- Instead of 'let expr', prefer (( ...
    make: *** [shellcheck] Error 1


rolfbjarne:
It looks like all the CI failed due to unrelated issues (download that fails, bots missing dependencies, etc).

